Issue #139, opened 2015-07-03
Cygwin exception


gurinderu:
OS: Windows 7 in Cygwin

Simple code: 
DefaultTerminalFactory factory = new DefaultTerminalFactory();
        factory.setSuppressSwingTerminalFrame(true);
        Screen screen = new VirtualScreen(new TerminalScreen(factory.createTerminal()));
        screen.startScreen();
        MultiWindowTextGUI textGUI = new MultiWindowTextGUI(screen);

        final BasicWindow window = new BasicWindow("Dialog test");
        textGUI.addWindow(window);

Exception in thread "main" java.io.IOException: Cannot run program "/bin/sh": CreateProcess error=2, The system cannot find the file specified
        at java.lang.ProcessBuilder.start(Unknown Source)
        at com.googlecode.lanterna.terminal.ansi.UnixTerminal.exec(UnixTerminal.java:313)
        at com.googlecode.lanterna.terminal.ansi.UnixTerminal.saveSTTY(UnixTerminal.java:289)
        at com.googlecode.lanterna.terminal.ansi.UnixTerminal.<init>(UnixTerminal.java:158)
        at com.googlecode.lanterna.terminal.ansi.UnixTerminal.<init>(UnixTerminal.java:106)
        at com.googlecode.lanterna.terminal.ansi.UnixTerminal.<init>(UnixTerminal.java:85)
        at com.googlecode.lanterna.terminal.ansi.CygwinTerminal.<init>(CygwinTerminal.java:46)
        at com.googlecode.lanterna.terminal.DefaultTerminalFactory.createCygwinTerminal(DefaultTerminalFactory.java:176)
        at com.googlecode.lanterna.terminal.DefaultTerminalFactory.createTerminal(DefaultTerminalFactory.java:90)
        at guru.grinder.Main.main(Main.java:23)
Caused by: java.io.IOException: CreateProcess error=2, The system cannot find the file specified
        at java.lang.ProcessImpl.create(Native Method)
        at java.lang.ProcessImpl.<init>(Unknown Source)
        at java.lang.ProcessImpl.start(Unknown Source)
        ... 10 more

avl42:
The root problem is, that UnixTerminal does more in its constructor than it should.
Therefore, all these alternatives coded in the CygwinTerminal just can't modify
base-class's (namely UnixTerminal's) construction.

To fix it, UnixTerminal ought to define a "bare" constructor to just set the fields,
but omit its Unix-specific trickery.

[...]

PR #140, opened 2015-07-07
Refactor UnixTerminal and CygwinTerminal.

avl42:
This is my first throw at untangling CygwinTerminal from UnixTerminal.
It is based on Issue #139 which shows, that CygwinTerminal inherits more
from its base-class than it can handle.

I don't have cygwin anywhere near hand, so I'd appreciate someone with
cygwin to try it and report on it.

PS: I did run some of the terminal test-programs for UnixTerminal, and they looked fine to me under Linux.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/terminal/ansi/CygwinTerminal.java
+++ b/src/main/java/com/googlecode/lanterna/terminal/ansi/CygwinTerminal.java
@@ -22,8 +22,6 @@ import com.googlecode.lanterna.TerminalSize;
 
 import java.io.*;
 import java.nio.charset.Charset;
-import java.util.Timer;
-import java.util.TimerTask;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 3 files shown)


mabe02:
Looks ok, I'd like to try it out at home first with Cygwin before merging

avl42:
Yes, under all circumstances, please test it with cygwin before merging to master!

Some remark: I wondered if stty's option -F would be available also for Unix, as that would easily allow skipping the shell, but while Linux's stty has it, unfortunately neither Solaris, nor AIX have it. It could still be possible, though, if stty's stdin could be redirected from /dev/tty in Java using the ProcessBuilder.

avl42:
I wasn't sure, if Java on cygwin can simply open "/dev/pts0" and pass it to stty.exe as stdin, so I made the changes ineffective for CygwinTerminal. Rather let stty.exe on cygwin open its device itself.

Eventually a better name might be found for UnixishTerminal  ;-)

One more thing:  If a UnixTerminal-instance is created with no streams specified, then, instead of System.in/out, shouldn't it rather open "/dev/tty" for read&write and use those streams? That would enable it to work like the "less"-utility: getting data from stdin, but interacting using the tty...

mabe02:
I've never thought of having it go directly to the tty, any particular benefit over using standard output?

mabe02:
Never got a chance to try it out for real with Cygwin, but I liked the change in general so I'll take it as it is. Will do some more investigation over weekend. One of the gotchas with running Java on Cygwin that hit me over and over is that the Java executable isn't compiled with cygwin so all the unix emulation is unavailable. Hopefully if all cygwin commands are on the $PATH it should still work to call them, but I do remember some strangeness when trying to change certain parameters (echo?) that just wouldn't take effect.

mabe02:
Renaming it to UnixLikeTerminal but not 100% happy with that either. Just that the usual AbstractUnixTerminal or BaseUnixTerminal or something like that doesn't appeal to me either. Open to suggestions...

Issue #134, opened 2015-06-29
Throw ArrayIndexOutOfBoundsException with 3.0.0-alpha6


GoogleCodeExporter:
```
The issues is InteractableLookupMap.getSize() assume it has 1 more rows. So
return new TerminalSize(lookupMap[0].length, lookupMap.length); will throw 
ArrayIndexOutOfBoundsException when it has 0 rows.
```

Original issue reported on code.google.com by `yuxuanch...@gmail.com` on 5 Jun 2015 at 9:07

avl42:
I don't know if something else was fixed to stop a zero-sized InteractableLookupMap from being created, but I think that the class itself should be immune to that problem. Thus the pull-request.

mabe02:
Yep, that's merged now

PR #145, opened 2015-07-10
make robust against ZERO size. (Issue#134)

avl42:
This should fix issue #134


avl42:
This should fix issue #134

PR #147, opened 2015-07-12
TextImage.toString, ListBox, IOSafeExtendedTerminal

avl42:
When debugging a TextGui program, it is most useful to be able to watch the Screen's backbuffer as it is drawn to, rather than only after a refresh.  This commit adds a toString() to BasicTextImage and ScreenBuffer, but also to AbstractScreen(which passes it on to its backbuffer).

So far these classes all had their default toString, which made examination of contents rather hard.

This toString() is still rather simple, as it ignores attributes, CJK and colours, but enough to still help.

The result is a bit odd for VirtualScreen, as its backbuffer contains some, but not all the visible stuff. I haven't yet found out, why that is so.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/graphics/BasicTextImage.java
+++ b/src/main/java/com/googlecode/lanterna/graphics/BasicTextImage.java
@@ -230,4 +230,17 @@ public class BasicTextImage implements TextImage {
             }
         };
     }
+    
+    @Override
+    public String toString() {
+        StringBuilder sb = new StringBuilder(size.getRows()*(size.getColumns()+1)+50);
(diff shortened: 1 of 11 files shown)


avl42:
Second commit: mostly riddance of compiler warnings.

avl42:
Third commit:  Introduces an "IOSafe and Extended" Terminal interface, as well as an "Extended" Adaptor, to make IOSafe-ness independent from Extended-ness.

This might be a preparation for eventually making SwingTerminal implement ExtendedTerminal, but whether this is ever done or not makes no difference for this commit.

mabe02:
Looks good!

Issue #138, opened 2015-06-29
Scrolling a Screen


cmiles74:
I'm using Lanterna 3.0.0-alpha6 and am trying to put together an irssi type interface. In my application the top-most portion of the terminal scrolls, the bottom line is reserved for user input. Presently when I need to scroll the content I redraw the top portion of the window in it's entirety, adding my new line to the bottom.

This works okay with the Swing terminal but performs poorly with an actual xterm (or Gnome Terminal) window. Is there a better way of scrolling text?

Thank you!

avl42:
This indeed doesn't seem to have been a design goal.  It is not easy, but also not impossible to work around it, if merely redrawing the terminal with the scrolled lines is really too slow.   I also had problems with flickering updates, until I started to call TerminalScreen.refresh(RefreshType.DELTA). Lanterna's default (per AbstractScreen.refresh()) is AUTOMATIC, which causes visible flickering when scrolling text.

Have a look at TerminalScreen's "refresh"-method: after deciding for refreshByDelta or refreshFull, these methods do the following: 1.) find differences between the back-buffer and the front-buffer (or an assumed empty screen in refreshFull), 2.) emit byte-sequences to the terminal to update it,  3.) copy the back-buffer to the front-buffer.

What you might need to do is derive a subclass from TerminalScreen, and add your own scroll-method that will apply the scrolling to both(front & back) buffers and send some specific sequences to the terminal to update it efficiently. You might need to do research for what sequences will do the scrolling the most efficient way and still work for most ANSI-like terminals.

If you do it, then pushing it to your github-fork and sending a pull-request to Martin (mabe02) would be nice. :)

mabe02:
I think in general the bottleneck comes down to the piece that actually pushed out the characters on the output stream, which (as avl42 is pointing out above) uses one of two strategies, either clear the whole screen and redraw everything from the top-left corner down to the bottom right corner (RefreshType.COMPLETE) or it compares the internal front- and back buffer and generates a sequence of characters and ASCI escape codes to turn the back buffer into the front buffer.

In your case, it's probably the COMPLETE refresh that's causing the flicker so I agree with avl42 that you should try the DELTA refresh. We may need to increase the threshold of when AUTOMATIC does the COMPLETE, possibly even deprecate AUTOMATIC and only use DELTA (unless explicitly asked to do COMPLETE) if it's a big source of flicker.

avl42:
[...]

PR #151, opened 2015-07-14
Scrolling

avl42:
Here is the PR for scrolling (Issue #138)

### DIFF
--- a/src/main/java/com/googlecode/lanterna/screen/AbstractScreen.java
+++ b/src/main/java/com/googlecode/lanterna/screen/AbstractScreen.java
@@ -191,7 +191,7 @@ public abstract class AbstractScreen implements Screen {
 
     @Override
     public void refresh() throws IOException {
-        refresh(RefreshType.AUTOMATIC);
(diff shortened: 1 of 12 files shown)


src/main/java/com/googlecode/lanterna/screen/AbstractScreen.java
```diff
@@ -191,7 +191,7 @@ public TextCharacter getBackCharacter(int column, int row) {
 
     @Override
     public void refresh() throws IOException {
-        refresh(RefreshType.AUTOMATIC);
+        refresh(RefreshType.DELTA);
```
mabe02: Rather, I think we should leave it at automatic and have the code handling automatic mode prefer DELTA.

avl42:
It really must be DELTA for the scrolling mechanism to work!

Otherwise it would compare the scrolled back buffer with the not-yet-scrolled front buffer and almost always come to the conclusion: full-redraw, which renders the whole scrolling feature pointless.

PR #153, opened 2015-07-15
Let users be able to use LinearLayout.Alignment

dant3:
I was trying to implement a toolbar similar to what nano does (see screenshot) and to do that wanted to use panels with horizontal linear layout filled on a panel. However I noticed that I'm unable to assign layout data to the action I was trying to put at the end of panel. This change fixes visibility of Alignment.

image


mabe02:
Yeah, that wasn't intentionally package private, thanks!

Issue #155, opened 2015-07-16
Crash on using Panel.setContent from ActionDialog callback


dant3:
I'm trying to update my ui from ActionDialog callback. What I'm doing is:
```scala
val window = new BasicWindow()
window.setComponent(createUi(None))
// ...
val dialog = new ActionDialogBuilder()
dialog.addAction(device.name, toRunnable(() => window.setComponent(createUi(Some(device)))))
dialog.build().showDialog(gui)
// ...
def createUi(device):Panel = Panels.vertical(component1(device), componentWhichMightCallDialogAbove)
//...
def toRunnable(fn: () => Unit):Runnable = new Runnable { override def run() = fn.apply() }
```

But then this Runnable action get's called I'm getting exception:

```
java.lang.NullPointerException
        at com.googlecode.lanterna.gui2.AbstractComponent.toBasePane(AbstractComponent.java:218)
        at com.googlecode.lanterna.gui2.AbstractComponent.toBasePane(AbstractComponent.java:218)
        at com.googlecode.lanterna.gui2.AbstractBasePane.getCursorPosition(AbstractBasePane.java:174)
        at com.googlecode.lanterna.gui2.MultiWindowTextGUI.getCursorPosition(MultiWindowTextGUI.java:181)
        at com.googlecode.lanterna.gui2.AbstractTextGUI.updateScreen(AbstractTextGUI.java:118)
        at com.googlecode.lanterna.gui2.MultiWindowTextGUI.updateScreen(MultiWindowTextGUI.java:144)
        at com.googlecode.lanterna.gui2.AbstractTextGUIThread.processEventsAndUpdate(AbstractTextGUIThread.java:61)
        at com.googlecode.lanterna.gui2.MultiWindowTextGUI.waitForWindowToClose(MultiWindowTextGUI.java:275)
        at com.googlecode.lanterna.gui2.AbstractWindow.waitUntilClosed(AbstractWindow.java:164)
        at com.googlecode.lanterna.gui2.dialogs.DialogWindow.showDialog(DialogWindow.java:26)
        at com.github.dant3.catlog.Main$.com$github$dant3$catlog$Main$$showDevicesList(Main.scala:58)
        at com.github.dant3.catlog.Main$$anonfun$createToolbar$1.apply$mcV$sp(Main.scala:44)
        at com.github.dant3.catlog.ui.Toolbar.com$github$dant3$catlog$ui$Toolbar$$handleKeyStroke(Toolbar.scala:20)
[...]

PR #159, opened 2015-07-17
Issue155 testcase reproducing the problem

dant3:
This is the testcase reproducing the problem I wrote about in Issue #155.
I made a situation which almost identical to the one I have in my code, it does not contains any custom components and it's written in Java ;)


avl42:
Martin, I suggest waiting a bit with this one, until Viacheslav either adds my proposed fix to it, or declines it.

dant3:
Closing this as test & fix was merged already.

PR #160, opened 2015-07-21
AutoCloseable for telnet classes

LTI2000:
Hi, I propose a change which lets you use TelnetTerminal and TelnetTerminalServer in try-with-resources blocks.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/terminal/ansi/TelnetTerminal.java
+++ b/src/main/java/com/googlecode/lanterna/terminal/ansi/TelnetTerminal.java
@@ -39,7 +39,7 @@ import java.util.List;
  * Also here: http://support.microsoft.com/kb/231866
  * @author martin
  */
-public class TelnetTerminal extends ANSITerminal {
+public class TelnetTerminal extends ANSITerminal implements AutoCloseable {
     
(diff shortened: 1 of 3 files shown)


dant3:
This breaks Java 1.6 compatibility

LTI2000:
We have Java 8, java 7 is already end of life

avl42:
It might be easy to upgrade Java on a PC, but it is far not so easy to upgrade it on certain Unix machines (e.g. AIX), whose vendors offer their own Java port.  I'm in the (relatively) "lucky" situation, that the machines to run my programs already have Java 7 on them, but I know it will take a long time, even now that Java 7 is officially "end of life", that I'll get to see a Java 8 (then probably on new hardware only).  I guess the situation isn't better for people stuck with Java 6 on some platform.

No doubt, there will some day be a point where dropping Java 6 is justified, but (imho) "AutoClosable" isn't such a point, not even ProcessBuilder.redirectInput.

One in a Java7/8 environment can create a subclass of those telnet classes, and have the subclass implement AutoClosable, if try-with-ressources is more than a convenience.

mabe02:
Yes, lanterna 3 will remain JDK 6 compatible

Issue #150, opened 2015-07-13
Issues with Lanterna Terminal Starting/Stopping


dant3:
I have a couple of inconvenience running Lanterna 3 app on OS X:

- then I start the app from Terminal, which is using `DefaultTerminalFactory` I always getting Swing terminal windows for my app. This doesn't feel nice. I expect app to run in terminal if I start it from terminal. I guess it has something to do with JVM flags.
- after running my app using `ANSITerminal` my mouse can't scroll terminal contents anymore. Instead it works as `↑` and `↓` arrow keys, making me close terminal to get rid of this effect. This is very inconvenient - mouse should work as it was before after I leaving the app. At least this is what happens if app exits due to exception. I guess there should be some shutdown hook to make it work as it used to before.

avl42:
not sure, if macos works that way, but on unix the decision is made upon existence of the DISPLAY environment variable. there is also a setting about preferrence of graphical or non-graphical terminal in the defaultterminalfactor, iirc.

You're right, exiting from private mode (the one where scrolling is inhibited) should be done in some exit-hook, even if the app terminates on an exception. Until that, you can try the command 'reset'. I think it also resets the private-mode.

dant3:
Yes, on OS X I can see DISPLAY defined in Terminal as well with value `/private/tmp/com.apple.launchd.EbruLstpGr/org.macosforge.xquartz:0`. From what I see in JDK sources the OS X doesn't checked for DISPLAY value, but instead, checked for awt toolkit property.
About the setting you referring to - will it make `UnixTerminal` preferred and fallback to `SwingTerminal` if `UnixTerminal` fails? What will happen if I'll allow user to start my app from Finder? From source code I see there is just `if (headless || preferUnixTerm) { doUnixTerminal(); } else { doSwingTerminal(); }` so it doesn't looks like there are some fallback handling happening.

`reset` does the trick. I think there is a way to preserve exception-dump in UncaughtExceptionHandler you can simply do what you have to do and finally just pass original exception to the original handler. I'll take a look at sources and see if I can implement it & create an appropriate PR.

avl42:
Sorry, no. The setting is merely to force terminal operation even when DISPLAY is set.
The toolkit-property behind "GraphicsEnvironment.isHeadless()" is probably initialized depending on presence of the "DISPLAY" env-var -- at least on unix-like platforms.

[...]

PR #162, opened 2015-07-21
fixes for issues #150 and #164

avl42:
In case of Ctrl-C or System.exit() or uncaught exception, real terminals should be completely reset: not just the stty-stuff, but also exit the private mode (which restores the visible cursor,...)

This is in response to issue #150 .


avl42:
second commit: fix for issue #164

avl42:
Thanks!

PR #165, opened 2015-07-26
Three fixes around scrolling

avl42:
As it was, automatic refresh mode would have spoilt scroll optimization...
I think after-all this approach is even better than my changing default to DELTA.

PS: follow-up to #151 and #138


avl42:
ok, wait a bit,   "three makes a charm"  ... third one arriving shortly...

avl42:
Ok, this third one is a corner case: When the screen is clear()ed, then scrolling the terminal makes no sense until next refresh...

avl42:
@mabe02, not noticed, yet, or just put off for later?

mabe02:
Sorry, I missed this one

PR #175, opened 2015-09-27
Simple examples

peterborkuti:
Hi,

Maybe these examples can be helpful to others. Feel free to reorganize/edit/etc or throw away them.

Thank you for your great library.

Péter Borkuti

### DIFF
--- a/src/examples/java/com/googlecode/lanterna/examples/DrawRectangle.java
+++ b/src/examples/java/com/googlecode/lanterna/examples/DrawRectangle.java
@@ -1,0 +1,40 @@
+package com.googlecode.lanterna.examples;
+
+import java.io.IOException;
+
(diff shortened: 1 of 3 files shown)


avl42:
Currently, there are many examples in the "test" directory-hierarchy.

I like that your examples are not only in their own source-tree, but also in their own (java-)package. This not only makes them easier to find, but also prevents them from accidentally using non-public API.

Of course, not all of the classes in *test* are example applications (some are JUnit-tests instead) and tests sometimes need to be in the same package as the library-classes they test.

The *examples* currently in *test*, however, ought to be moved to where the new examples are placed. (@mabe02)

mabe02:
Agreed, better to sort them separately in a clear package not (mis-)using package private api

PR #180, opened 2015-11-03
Add shiftDown support to KeyStroke

avl42:
This is a first step towards supporting shifted cursor keys, function keys, home,end,pgup,pgdn.

Shift-flag is specifically **not** supported for character KeyStrokes. E.g.: "Shift-6" is '^' for English keyboards and '&' for German ones. Shift-Tab is still possible, but should be mapped to ReverseTab.

Next step will be a new CharacterPattern implementing class, that will replace  most of the individual BasicCharacterPatterns in the DefaultKeyDecodingProfile, and will support any combination of alt,ctrl,shift for some special keys, while avoiding to specify all 8 variants for each of them.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/input/DefaultKeyDecodingProfile.java
+++ b/src/main/java/com/googlecode/lanterna/input/DefaultKeyDecodingProfile.java
@@ -36,104 +36,17 @@ public class DefaultKeyDecodingProfile implements KeyDecodingProfile {
             = new ArrayList<CharacterPattern>(Arrays.asList(
                             new CharacterPattern[]{
                                 new BasicCharacterPattern(new KeyStroke(KeyType.Escape), ESC_CODE),
-                                new BasicCharacterPattern(new KeyStroke(KeyType.ArrowUp), ESC_CODE, '[', 'A'),
-                                new BasicCharacterPattern(new KeyStroke(KeyType.ArrowUp, false, true), ESC_CODE, ESC_CODE, '[', 'A'),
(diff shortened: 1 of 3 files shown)


avl42:
The current DefaultKeyDecodingProfile seems to assume, that Alt-Fn keys are sent with an extra Escape before the normal Fn sequence.

I haven't found any terminal yet, that did it this way.  Instead they all produced a sequence like the one for Ctrl, just with a different second number. In case it turns out, that there really are terminals that send an Escape-prefix also for non-letter keys, then there are three lines to un-comment in EscapeSequenceCharacterPattern.java, to support even those sequences.

Linux (console) F[1-5] keys do not follow the new sequence-pattern, so they're still left in DefaultKeyDecodingProfile as individual BasicCharacterPattern-entries.

I tested these changes with gnome-terminal and JuiceSSH (some terminal-App on my tablet) and I'd appreciate others testing with cygwin, putty and/or other terminal emulators.

@mabe02, I'd appreciate any reply, just so I know you didn't miss it - also for the other PR (about waiting for complete sequences, which I consider finished now)

mabe02:
Looks good to see all those patterns go!

avl42:
quick note: @mabe02, I noticed that putty actually uses esc-prefixing for "Alt-special keys". 

Can you find&fix those three lines to un-comment and test with putty? (in method matches, after second case-label) - otherwise I'll do it next week.

mabe02:
I think this is what you want?
6956b1b1d990ab2d8ad3c9d6a3919e4164379e14

avl42:
Mostly. Please, see my comments on the 6956b1b commit

PR #186, opened 2015-11-15
Update tests

ben-dale:
A couple of changes to tests accidentally slipped in when https://github.com/mabe02/lanterna/pull/185 was merged in. This PR reverts those changes.

PR #187, opened 2015-11-16
Restore tests to former glory (ben-dale)

avl42:
This is a redo of ben dale's #186 with the only change, that it contains just a single commit of those two files, and avoids the merge-commit, leaving the resulting version tree just a tiny bit leaner.

Not sure if that was even worth it, but well...  now mabe02 has the choice :-)


ben-dale:
This is the correct PR, thanks @avl42!

mabe02:
Thank you!

PR #191, opened 2015-12-04
I18n handling proposal

SilverYoCha:
Only the label of button component has been taken into account.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/Config.java
+++ b/src/main/java/com/googlecode/lanterna/Config.java
@@ -1,0 +1,21 @@
+package com.googlecode.lanterna;
+
+import com.googlecode.lanterna.bundle.BundleLocator;
+
+import java.io.File;
+
+/**
+ * Centralized configuration class.
+ * @author silveryocha
+ */
+public class Config {
+
+  /**
+   * Gets the locale user directory into which the lenterna data will be read or save.<br/>
+   * {@link BundleLocator}, for example, can read multilingual property files from this space.
+   * @return
+   */
+  public static File getLocaleStorageFile() {
+    return new File(System.getProperty("user.home"), ".lanterna");
+  }
+}
--- a/src/main/java/com/googlecode/lanterna/I18n.java
+++ b/src/main/java/com/googlecode/lanterna/I18n.java
@@ -1,0 +1,28 @@
+package com.googlecode.lanterna;
+
+import com.googlecode.lanterna.bundle.MultilangLocator;
+
+import java.util.Locale;
+
+/**
+ * This class permits to get i18n translations about lanterna components by an easy way.
+ * @author silveryocha.
+ */
+public class I18n {
+
+  private static final MultilangLocator multilang = MultilangLocator.getMultilang("lanterna-ui");
+
+  /**
+   * Gets the message with parameters
+   */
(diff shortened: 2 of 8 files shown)


avl42:
1) It seems to me like the approach is to just "slap-on"-translate whatever some mono-lingual programmer wrote in the application. Imho, that is a bad approach. E.g.: "open" in English is ambiguous and translations might be different depending on whether it means the command to open something, or as a status-report of something that is(or isn't) currently open. I guess there are more words that cannot sensibly be translated without context. Usually i18n-applications therefore use symbolic names (that bring some context to the wording), and are then translated to English just like to any other language.

  2) Having Button actually replace it's label with the translation is imho not a good idea. There should always be a separation of internal strings and localized messages. Mixing these leads to disaster. An application might e.g. save a current label, place a new temporary label, then restore the old one - but the old saved one was already translated, and might end up chain-translated...

Apart from that, it will be Martin to decide if adding localization helper classes to Lanterna brings any advantages to whatever localization support Java itself already offers.

SilverYoCha:
Thank you for your analysis.

As a result, and after reading your feedback, I think I did not take enough time to choose the right semantic for my class names. E-g, instead "I18n" maybe LocalizedBundle would be more appropriate. So, you are right on your first point of scepticism.

And you are right also about your second one! I totally agree with : "There should always be a separation of internal strings and localized messages". Initially, I wanted to do something for MessageDialog where the button labels was the name of one of MessageDialogButton enum attribute. But it was not enough, some dialog buttons where not translated. And I took the bad way, that is true, to get the translations performed all the time (for button component)...

Finally, I refactored my work and simplified it.
If you are interested in, I can create a new pull request, or push it over my current one.
And if you are not, I will just go on...

As I am glad to use the work of all lanterna contributors, I wanted to do something in return...

Thank you for your pertinent feedback.

avl42:
I wasn't really aware of the Dialogs and that they do contain hardcoded (not user-replaceable) English words...  That is indeed a spot that should be improved in Lanterna.

I think, i18n should be mainly the application's responsibility (and not so much this library's), but there needs to be a *way* for the application to actually choose it's own texts. As a simple (not necessarily optimal) example, the MessageDialog(Builder) could accept an EnumMap<MessageDialogButton,String>-typed parameter, that would map each MessageDialogButton to a string of user's language to be used as visible label. Alternatively, it might add some prefix like "MessageDialog_" to each of the labels and try to look that up in some standard I18n-helper class.  Any way, the application would internally still "see" the original MessageDialogButton identifiers, but the user would see (ideally) his own language's texts.

Also, the chosen Language should not be stored in a *static* field, because Lanterna can be used (e.g. through it's TelnetTerminalServer) by different users (with maybe different languages) in a single Java VM.

I'm very curious about your update, and if it is quite different from this one, then I'd think that a new PR with just the relevant commits for the new approach would make sense.

Thanks for spotting a point where to further improve Lanterna!

avl42:
PS: I haven't yet done I18N on Java.   I saw it done in some C++ project using fancy macros and a website to have programs translated by volunteers (https://transifex.com), but I don't really know how this is usually done in Java.  There is a tutorial for how to pick translated texts https://docs.oracle.com/javase/tutorial/i18n/resbundle/index.html  but I haven't yet completely read it myself...

PR #192, opened 2015-12-14
Localization proposal for short labels (OK, Yes, No, etc.)

SilverYoCha:
This PR is the result of the discussion on https://github.com/mabe02/lanterna/pull/191.

I'm sorry to propose it after one week and more while the refactoring was already done...

As the proposal of my first PR, the mechanism uses the built-in one provided by Java. But this time, it is more simple and righter.

For now, only the button labels are taken into account and there are two handled languages:
- the default one, English(lanterna-ui.properties)
- the French one (lanterna-ui_fr.properties)

To get another handled language, the linked lanterna-UI_XX.properties file must be added.
For example, to handle German language, lanterna-ui_de.properties must be created.

I have read your post-scriptum about https://transifex.com, and I think that it is a good way to get translations, but I'm not sure that it is a good thing to integrate it directly into lanterna as it is clearly WEB oriented. However, I'm not one hundred per cent sure but it seems possible to make a connection between a GitHub project to their WEB APIs. Maybe the bundle property files could be filled by this way...

I hope this PR can be interresting, or at least, may make other ideas emerge about the subject of Localization.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/bundle/BundleLocator.java
+++ b/src/main/java/com/googlecode/lanterna/bundle/BundleLocator.java
@@ -1,0 +1,58 @@
+package com.googlecode.lanterna.bundle;
+
+import java.security.PrivilegedAction;
+import java.text.MessageFormat;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import java.util.ResourceBundle;
+
+/**
+ * This class permits to deal easily with bundles.
+ * @author silveryocha
+ */
+public abstract class BundleLocator {
+
+    private final String bundleName;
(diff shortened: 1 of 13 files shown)


avl42:
This one is certainly much better!

There are a few minor things that I'll annotate at their spots.

I would even consider adding a new constructor to Button (and maybe Label, too) that takes a LocalizedString (and just calls `this(ls.toString(),...)`, so one could skip the `.toString()` at call-site.

mabe02:
Looks ok, I'll review this in more details over the weekend and try to merge it.

mabe02:
I'm just going to do one small adjustment to one of the class names, otherwise great stuff!

Issue #198, opened 2016-01-20
MultiWindowTextGUI ignoring Window.isVisible()


nathanws:
Looking through the interface code for `Window`, it contains a method `isVisible()`. The comments for that method say that it could be used by `WindowManager`s to decide if the `Window` should be drawn or not. 

In `DefaultWindowManager` (the only implementation of the interface), it doesn't handle if the `Window` should be drawn or not. 

Is this a planned feature? Or something that was changed with the move to version 3.0? Or am I missing something?

To temporarily work around this, I copied the code from `MultiWindowTextGUI` to a new class and modified the `updateScreen()` and `drawGUI(TextGraphics)` to do a check against `Windows.isVisible()`. It works great.

Normally, I would create a class that extends `MultiWindowTextGUI` and only override the appropriate methods. The problem with doing this, is that the fields `postRenderer` and `virtualScreen` do not have getters. 

I suggest either adding the getters, or having `MultiWindowTextGui` use `Window.isVisible()` to determine if the `Window` should be drawn.

I can do either and send a pull request if you like. Also, is there a roadmap for development? I would really like to contribute.

avl42:
To me it seems that your change makes sense as described.

I'd suggest you to make a PR, so next time Martin is here he can already merge it in.

I'd also include the getters into the PR (even if you wouldn't immediately need them any more after the visibility-check is built in) - it might still be useful later, if anything else eventually needs adaptation.

bah-lingt:
Note that AbstractWindow sets the visible flag to false in the constructor, and I can't seem to find a setter for this variable. With this change, all new windows will be invisible by default.

nathanws:
That's a good point, something I missed. I had a class that extended `AbstractWindow`, and I overrode the `isVisible()` method.

Since `AbstractWindow` is already mutable, I don't see a reason why a setter for `visible` can't be added there.

mabe02:
This should be all merged now, including the setVisible(..) method

PR #201, opened 2016-01-25
MultiWindowTextGUI respects Window.isVisible(). Exposed VirtualScreen…

nathanws:
… and WindowPostRenderer in WindowBasedTextGUI.

As mentioned in Issue #198, I made `MultiWindowTextGUI` respect the `Window.isVisible()` property when drawing the GUI and when updating the screen.


src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
```diff
@@ -425,6 +429,16 @@ public BasePane getBackgroundPane() {
     }
 
     @Override
+    public VirtualScreen getVirtualScreen() {
+        return virtualScreen;
+    }
```
mabe02: We should just Screen here, not VirtualScreen

mabe02:
Thank you for the fix, looks good

PR #203, opened 2016-02-04
argument to stopScreen to control flushing of input

avl42:
I'd like to be able to avoid flushing of the input buffer on exit.  This PR might just be a start - next step would be getting rid of the Bufferedreader in InputDecoder.


mabe02:
Sure, let's call it flushInput just to emphasise

Issue #210, opened 2016-02-22
Application crashes when resizing window (bad "out of bounds" check in TerminalTextUtils)


nikiroo:
Hello,

I'm trying out Lanterna3.
I've got a few issues, but one of them is easy to debug and solve so I thought it may help you to know about it.

In some cases, resizing the terminal (e.g., xterm, xfce4-terminal...) causes a crash in Lanterna 3.
It comes from the line:
 ```
while(Character.isSpaceChar(row.charAt(characterIndex)) && characterIndex < row.length()) {
```
at:
```
com.googlecode.lanterna.TerminalTextUtils.getWordWrappedText(TerminalTextUtils.java:237)
```

Notice that the "index < length" check is done AFTER the index is used.

Inverting the 2 checks does indeed fix this specific issue.

Thanks,
Niki

avl42:
Good catch! Thanks for reporting.

And it turns out, it's not the only thing...
- the loop at line 228 also has `characterIndex > 0` as last condition. Here it won't hurt, however, except for unnecessarily testing 0'th character for space'ness or cjk'ness.
- in line 236 it cuts the string at maxWidth, but maxWidth is a "physical" width, not a number of characters. If the line has CJK chars, then some chars will end up hidden behind right edge.
- variable spaceCharsToSkip is either unused, or whatever were the plans for it aren't completed, yet.
- finally, the original point of this issue.  (It only occurs when the wrapped part of the line is whitespace-only.) it has to be decided if we even want to wrap it then, or just cut it.

Unless you @nikiroo or @mabe02 are faster to it, I'll probably try to fix it and make a PR later this day (I don't have eclipse near hand at the moment).

avl42:
Small correction of my previous message: the loop at 228 doesn't really consider the first character.
it should really check for characterIndex >= 0 (greater-equal, not just greater), first. That way, characterIndex will stay at 0 if 0'th character is a space or cjk, and it will be -1 if there really is no space/cjk in the row.  Line 234 then becomes `if (characterIndex < 0) {` .

avl42:
And not to forget the case, where maxWidth == 1 and there is a CJK at the start of the row.

Would it be moved to next row forever?  Whoever does it, please also add a testcase for that.

mabe02:
Thanks everyone, I've merged the PR

PR #211, opened 2016-02-23
Fix bugs mentioned in Issue#210 and comments

avl42:
Ok, this is the final list of bugs identified and fixed for issue #210:
- Never break at char position 0: always leave at least one char before a line break (to prevent infinite loop)
- Also allow to break immediately after a CJK character
- Where no nice word-wrap is possible break by available space. (Just in case we eventually get non-breaking double-width chars, later.)
- Fix StringIndexOutOfBoundsException
- cut off trailing white-space after line break, rather than putting an empty string in a new line.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/TerminalTextUtils.java
+++ b/src/main/java/com/googlecode/lanterna/TerminalTextUtils.java
@@ -224,26 +224,41 @@ public class TerminalTextUtils {
             }
             else {
                 //Now search in reverse and find the first possible line-break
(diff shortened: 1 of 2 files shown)


avl42:
Oups, I forgot the testcase for trailing whitespace:
```
        assertEquals("Incorrect word wrapping - don't add a new line for only white space",
                Arrays.asList("abc "),
                TerminalTextUtils.getWordWrappedText(4, "abc       "));
```

mabe02:
Looks good, thanks!

PR #241, opened 2016-06-18
FileDialog : NullPointerException when directory is null

thlabbe:
Hi,
I get a NullPointerException when the directory is null :  selecting the parent from a root directory NPE and application is locked
( on Windows I would excpect the dirve's list ... I didn't had a try on other OS ... maybe no parent for root directory ?? )

With this quick and dirty fix : no more NPE but the fileDialog becomes empty. 
No Idea how to retrieve the Drive list just for windows.

Thomas Labbé


thlabbe:
here comes a clever fix : preventing the directory to be null

avl42:
haven't looked into lanterna's dialog, yet, but Java's way to deal with various platform's roots is java.io.File.listRoots()

thlabbe:
fine, thank you !
I will be able to change between drives

2016-06-19 2:42 GMT+02:00 mabe02 <notifications@github.com>:

PR #242, opened 2016-06-19
FileDialog.java : adding other file System roots available

thlabbe:
When filesystem root is reached other available roots are shown ( on windows other roots are the drives )


mabe02:
I've tried it out and it seems to work, thank you!

Issue #239, opened 2016-06-02
Request: spannable text?


jslater89:
It would be great to be able to set bold or underline on certain spans of text in a label, so I could emphasize certain parts of it without having to combine labels.

jslater89:
Alternately, implementing some ANSI escape codes in the Swing terminal might be easier, and that's probably something I'm capable of doing on my own.

mabe02:
I can see how this would be useful, but unfortunately, it's going to be pretty low on the priority list as it's a quite large task. Maybe you can override label or create a custom component to make it look the way you like?

avl42:
This consist of three parts:

1. define a markup language. maybe some subset of html, or something simpler.
2. write a renderer that would render the markup language into a TextImage.
3. let components of lanterna accept a TextImage just like they accept a String now, and just do the simple parts of re-displaying that TextImage instead of drawing a String.

If lanterna could add just the "3.", then this would already help some users.

avl42:
Finally, that one TextImage could be changed to a sequence of TextImages (one for each word) to allow for word-wrapping lateron. Some special values in that sequence could be used for forced line breaks and other fancy formatting hints.

PR #245, opened 2016-07-11
Framework for handling ANSI escape sequences for selective styling of running text

jslater89:
Implements underline. See also issue #239.

Should have minimal performance impact outside of text which uses escape sequences.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/TerminalTextUtils.java
+++ b/src/main/java/com/googlecode/lanterna/TerminalTextUtils.java
@@ -36,6 +36,54 @@ public class TerminalTextUtils {
     private TerminalTextUtils() {
     }
 
+    public static String getANSIControlSequenceAt(String string, int i) {
+        String controlSequence = null;
+
+        char character = string.charAt(i);
+        if (character == 0x1B) { // escape
+            if (string.length() > i + 2) { // Control sequences require a minimum of two characters
+                char possibleCSI = string.charAt(i + 1);
+                if (possibleCSI == 0x5b) { // left bracket opens a control sequence
+                    for (int j = i; j < string.length(); j++) {
+                        char controlChar = string.charAt(j);
+                        if (isCharCSITerminator(controlChar)) {
+                            controlSequence = string.substring(i, j + 1);
+                            break;
+                        }
+                        else if (j > i && controlChar == 0x1b) {
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        return controlSequence;
+    }
+
+    public static boolean isCharCSITerminator(final char c) {
+        return c == 'A'
+                || c == 'B'
+                || c == 'C'
+                || c == 'D'
+                || c == 'E'
+                || c == 'F'
+                || c == 'G'
+                || c == 'H'
+                || c == 'J'
+                || c == 'K'
+                || c == 'S'
+                || c == 'T'
+                || c == 'f'
+                || c == 'm'
+                || c == 'i'
+                || c == 'n'
+                || c == 's'
+                || c == 'u'
+                || c == 'l'
+                || c == 'h';
+    }
+
     /**
      * Given a character, is this character considered to be a CJK character?
      * Shamelessly stolen from
--- a/src/main/java/com/googlecode/lanterna/graphics/AbstractTextGraphics.java
+++ b/src/main/java/com/googlecode/lanterna/graphics/AbstractTextGraphics.java
@@ -273,8 +273,21 @@ public abstract class AbstractTextGraphics implements TextGraphics {
         }
         string = tabBehaviour.replaceTabs(string, column);
         int offset = 0;
+
+
+        EnumSet<SGR> localModifiers = activeModifiers.clone();
         for(int i = 0; i < string.length(); i++) {
             char character = string.charAt(i);
+
+            String controlSequence = TerminalTextUtils.getANSIControlSequenceAt(string, i);
+            if(controlSequence != null) {
+                localModifiers = updateLocalModifiers(controlSequence, localModifiers);
+
+                // Skip the control sequence, leaving one extra, since we'll add it when we loop
+                i += controlSequence.length() - 1;
+                continue;
+            }
+
             setCharacter(
                     column + offset,
                     row,
@@ -282,7 +295,7 @@ public abstract class AbstractTextGraphics implements TextGraphics {
                             character,
                             foregroundColor,
                             backgroundColor,
-                            activeModifiers.clone()));
+                            localModifiers.clone()));
             
             if(TerminalTextUtils.isCharCJK(character)) {
                 //CJK characters are twice the normal characters in width, so next character position is two columns forward
@@ -343,5 +356,32 @@ public abstract class AbstractTextGraphics implements TextGraphics {
 
     private TextCharacter newTextCharacter(char character) {
         return new TextCharacter(character, foregroundColor, backgroundColor, activeModifiers);
+    }
+
+    private EnumSet<SGR> updateLocalModifiers(String controlSequence, EnumSet<SGR> localModifiers) {
+        char controlCodeType = controlSequence.charAt(controlSequence.length() - 1);
+        controlSequence = controlSequence.substring(0, controlSequence.length() - 1);
+        controlSequence = controlSequence.replace((char) 0x1b, '[');
+        controlSequence = controlSequence.replaceAll("\\[", "");
+        String[] codes = controlSequence.split("\\;");
+
+        EnumSet<SGR> newModifiers = localModifiers.clone();
+
+        if(controlCodeType == 'm') { // SGRs
+            if(codes.length == 0) { // Reset to default
+                newModifiers = activeModifiers.clone();
+            }
+            else {
+                for (String s : codes) {
+                    if (s.equals("4")) {
+                        newModifiers.add(SGR.UNDERLINE);
+                    } else if (s.equals("24")) {
+                        newModifiers.remove(SGR.UNDERLINE);
+                    }
+                }
+            }
+        }
+
+        return newModifiers;
     }
 }


mabe02:
Why only underline?

jslater89:
It was the one I needed. :P 

More broadly, I didn't want to make too many potentially breaking changes. I may go back and add some more styles as I have the time.

src/main/java/com/googlecode/lanterna/TerminalTextUtils.java
```diff
@@ -36,6 +36,54 @@
...
 
+    public static String getANSIControlSequenceAt(String string, int i) {
+        String controlSequence = null;
+
+        char character = string.charAt(i);
+        if (character == 0x1B) { // escape
+            if (string.length() > i + 2) { // Control sequences require a minimum of two characters
+                char possibleCSI = string.charAt(i + 1);
+                if (possibleCSI == 0x5b) { // left bracket opens a control sequence
+                    for (int j = i; j < string.length(); j++) {
```
avl42: I think, this loop could start with `j=i+2`

src/main/java/com/googlecode/lanterna/graphics/AbstractTextGraphics.java
```diff
@@ -344,4 +357,31 @@ public TextGraphics newTextGraphics(TerminalPosition topLeftCorner, TerminalSize
     private TextCharacter newTextCharacter(char character) {
         return new TextCharacter(character, foregroundColor, backgroundColor, activeModifiers);
     }
+
+    private EnumSet<SGR> updateLocalModifiers(String controlSequence, EnumSet<SGR> localModifiers) {
+        char controlCodeType = controlSequence.charAt(controlSequence.length() - 1);
+        controlSequence = controlSequence.substring(0, controlSequence.length() - 1);
```
avl42: Why not just controlSequence.substring(**2**, controlSequence.length() - 1) and avoid this rather expensive two-steps elimination of Esc and `[` ?

src/main/java/com/googlecode/lanterna/graphics/AbstractTextGraphics.java
```diff
@@ -273,16 +273,29 @@ public TextGraphics putString(int column, int row, String string) {
         }
         string = tabBehaviour.replaceTabs(string, column);
         int offset = 0;
+
+
+        EnumSet<SGR> localModifiers = activeModifiers.clone();
```
avl42: Since `updateLocalModifiers` doesn't change its parameter but instead works on a cloned Set, this first `.clone()` seems unnecessary.

avl42:
I've inline-commented a few minor improvements, but there are also larger ones (in terms of lines to change), which I'll first describe here roughly:

iirc, CSI sequences shouldn't have anything but digits and semicolons before their terminating controlCodeType. So, instead of searching for one of a longish list of characters, we could just scan over digits and semicolons, and anything else would become the controlCodeType. A sequence like `Esc[1;2~` would also be treated as some dummy Escape sequence with no effect - is that a problem or rather a win?

Also, I'd have getANSIControlSequenceAt() or a new (differently named) TextUtility method return just the length of the sequence. Some callers might only want to call it to see how many of following characters would be actually invisible, in order to calculate best line-break positions. And code which really needs the escape sequence as a String can easily do a text.substring(i,i+len) afterwards.

If you're not sure about it, I could create a PR to clarify what I'm thinking.

A second problem with the whole approach would be multi-line rich-texts: if a text starts with the underline sequence, then only the first line (putString() call) would see the sequence, and there seems to be no way for the caller to know, whether some attributes were changed. We might need a variant of putString that would have a means to tell its caller about changed attributes, so the caller can use this information for a next call to putString (e.g. for the next line of flowing multi-line text).

jslater89:
Good points all around. When next I have some time, I'll implement the smaller ones. I *think* you're right about CSI sequences, but I'd want to look and make sure there wouldn't be any perverse effects.

As for multi-line marked-up texts, I hadn't thought of that. Would it work to push localModifiers (or a list of csiModifiers, which would be merged with the active modifiers prior to rendering) out into the TextGraphics as a field?

avl42:
Another thing worth considering is, that while at the moment only underline is implemented, we might later want to support not only all attributes, but also changing colors - essentially all the stuff a TextCharacter can hold besides the char.

I think the TextGraphics is already a fine "container", and putString* (upon seeing a sequence) should just update the TextGraphics' attributes directly, instead of some localModifiers copy.

Maybe we should still have a separate putStringEx to interpret any sequences, while plain putString would just take characters literally as before, just so some "tainted" String wouldn't be able to hide important text following it (like by setting fg color to bg, once such is possible).

mabe02:
I like the idea of having a second version of putString that decorates the text

jslater89:
> I think the TextGraphics is already a fine "container", and putString* (upon seeing a sequence) should just update the TextGraphics' attributes directly, instead of some localModifiers copy.

I don't think this is correct behavior, though: say we've set a piece of text to bold using Lanterna functionality. Say we change color and underline status using CSI codes, then we add a CSI RESET code (to undo the color and underline stuff we've done). That would also wipe out the Lanterna-set bold status if we don't track running styling via CSI separate from styling done through the Lanterna API.

It also strikes me that, because of issues like some terminals not supporting e.g. BOLD_OFF, we should probably also strip CSI sequences that we handle from the text before passing it along.

avl42:
I think I understand what you mean. I just don't entirely agree ;-)

I'm thinking of a compromise between allowing fancy cell-formattings
controlled by escape sequences, while not making it overly sophisticated.

It is rather easy to obtain a new TextGraphics tg2 from current tg1 to
write some strings then forget tg2 and go back to use tg1 afterwards, which
will still have the original attributes and colors.

Otoh, controlling two layers of attributes (the current set, and the
modifiers on top of it) makes things pretty complicated - both inside
Lanterna and in the user-visible API.

avl42:
Devices not supporting e.g. BOLD_OFF aren't an issue, either, because these
escape codes are never directly sent to the terminal.  They're only
interpreted to set the style for next TextCharacters.  Assembling of
sequences for the actual device is TerminalScreen.refresh()'s job.

avl42:
> Otoh, controlling two layers of attributes ...

This also raises other questions, like if the TextGraphics was already set
to underline, and the text contained the underline-off sequence: does that
mean that the following characters should be plain (not underlined), or
that underline is just removed from the stacked style-layer, leaving the
next characters underlined. Would there need to be an extra sequence for
"even if default was already underlined, then please temporarily inhibit
it, until next sgr0")

For this, each flag would need to become tri-state: default (as set per
TextGraphics-API), override-to-set, override-to-unset.  That's what I meant
by "pretty complicated."

jslater89:
> It is rather easy to obtain a new TextGraphics tg2 from current tg1 to write some strings then forget tg2 and go back to use tg1 afterwards, which will still have the original attributes and colors.

This presupposes that you're working on the text-graphics level. I'm at the TextGUI level. I have a window which presents a scrolling view of status messages, some part of which I want to underline selectively. Unless I'm misunderstanding, this is not really feasible to do without some sort of markup.

> For this, each flag would need to become tri-state: default (as set per TextGraphics-API), override-to-set, override-to-unset. That's what I meant by "pretty complicated."

Why bother with another layer of formatting for SGR markup, though? I think that's an unwarranted complication. There isn't any reason to make SGR support robust enough to do all the styling you would want to do in an application, because Lanterna already has support for almost every case. I think it's perfectly reasonable to leave SGRs for the exceptional cases. For instance, the only reason I added this feature is because there was no other way to mark up a small part of the text in a larger TextGui element for different styling.

avl42:
I'm pretty confused now - there might have been misunderstandings...

We're talking about interpretation of esc-sequences in strings passed to putString.  At the moment, the effect of these sequences gets stored in a local copy of activeModifiers, namely localModifiers, and this is lost when putString finishes.  Now the question is: where shall we store it, so subsequent putString calls could pick up  the latest set of modifiers as modified by previous esc-sequences.

I see four options for that:

1. leave it as it is: sequences' scope is limited to a single putString call.

2. make an overload of putString with an extra parameter pointing to a mutable class - any styles set from esc-sequences in the string would be stored there.

3. Let the TextGraphics store two sets of styles: the one that is set directly through setters, and the other one that gets set by esc-sequences.  It's somewhat complicated to define how these two style sets will be combined to determine the style for subsequent characters.

4. Let the esc-sequences modify the one and only set of styles of the TextGraphics. The simplest, but it will not allow a sequence to reset to any previous state - the price of simplicity.

mabe02:
Maybe we should limit this to TextGUIGraphics? It seems like this will be very complicated to add to the TerminalTextGraphics implementation (which isn't 100% implementing the TextGraphics interface, sure, but still) and the compelling use case for doing text markup in the GUI layer doesn't really apply to the Terminal layer. I guess one could make a case for adding it at the ScreenTextGraphics level too.

With that in mind, I think option (1) is the best, the effects only take effect within the scope of a single putString call, although possibly under a different name as a separate method, so putString would do no escape sequence interpretation. Possibly make a new interface ("MarkupTextGraphics"?) exposing this.

Thoughts?

jslater89:
>Let the TextGraphics store two sets of styles: the one that is set directly through setters, and the other one that gets set by esc-sequences. It's somewhat complicated to define how these two style sets will be combined to determine the style for subsequent characters.

This is the bit that I dispute. I think it's reasonable to say that the Lanterna set should apply first, and any SGRs applied after. That said, I think mabe02 is correct as far as location goes: markup is only really a necessary concept in the GUI layer, and in that context, I don't think it needs to persist across multiple putStrings.

I'd be more than willing to take a crack at cleaning things up, but I'm not all that familiar with Lanterna's internals, and I'm not sure how best to go about, say, adding the functionality at the GUI layer but not below it.

mabe02:
Please have a look at the commit above, I've moved out the CSI-parsing part to a separate method call. I had a couple of different ways of approaching this initially and it was hard not to over-design it.

Unfortunately, this features opens up a huge rabbit hole of extra features you would need in order to really be able to use it properly. For example, there's no method to measure how long a string, minus the escape codes, is. Or how to print a sub-string given start/end character indexes that are not escape-code aware. Or given an index in a string with escape codes and given an initial SGR state, what is the color/styles at character position X? Or simply just map a string with escape codes to one without.

Also, if we get CSI 0 m, do we really reset everything or do we reset to the entry state of the method call? What about the state of the TextGraphics after the call, does it go back to the previous state at the start of the method or does it keep whatever the last state of the string was?

jslater89:
The huge rabbit hole of features occurred to me, and I'm not sure how to most gracefully handle the cases you've posited. Do they come up often?

As far as resetting, I'd expect CSI/SGR modifiers in the text to overlay modifiers passed in/themed with Lanterna, and resetting to remove the overlay, at least at the text graphics level.

As far as the commit goes, how would one go about setting, say, Label to use putCSIStyledString instead of putString?

mabe02:
Maybe spanning text is easier after all? Although I think we'll end up in a similar situation,

mabe02:
Okay, let's take a step back. What was your use case for this again?

jslater89:
In [this screenshot](http://imgur.com/a/KtOtz), the window at the right of the screen has two pertinent components visible components: one label ('Status'), and a second, multi-line label serving to display a buffer of past status messages, extending between the Status label and the Clock label. It's useful for me to be able to underline/style arbitrary bits of past status text to highlight them and provide separation, like 'Last move'.

avl42:
So, is this window at the right a single widget for which you have
overridden the drawComponent method,  or is it a Container into which you
add a number of single label-widgets?  is each line a label, or is each
block of information a label widget?

I do think, that labels (and some other widgets) should support
style-markup.  The markup, however, doesn't fit into the level of
TextGraphics.putString().

Instead, there should be a "TextWriter" on top of the TextGraphics, and the
TextWriter should keep track of "current position" and might even include
word-wrapping, too.

Small widgets would then create a TextWriter (which would pick default size
and current style from its backend TextGraphics), and pass the whole
content to it. The TextWriter then would deal with style markup,
word-wrapping, embedded line-feeds and even tabs (currently still in
TextGraphics).)

Larger widgets would split their content into chunks and pass each chunks
to TextWriter method, and after each chunk cache the resulting style.  If
it later needs to redraw only the bottom of the text, it only needs to
catch up at a chunk-boundary and doesn't need to re-render all the text.

jslater89:
Each block of information is a label widget: the 'status' and 'clock' labels are simple labels, and the object containing the actual status text subclasses label. (It handles line wrapping and scrolling.)

avl42:
I'd like to link to my new PR #273 where I implemented the ideas to some degree.
There might still be issues with it, but I think it's heading in right direction.

PR #258, opened 2016-08-23
Fix junit dependency scope

nicoulaj:
junit dependency should have `test` scope.


mabe02:
Weird that I didn't notice that all these years...

PR #260, opened 2016-08-24
Fixed scrollbar theming

ekoutanov:
Calling onAdded ensures that the scrollbar has its parent set. Without
it, scrollbar’s default renderer will revert to the default theme.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/table/DefaultTableRenderer.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/table/DefaultTableRenderer.java
@@ -357,6 +357,7 @@ public class DefaultTableRenderer<V> implements TableRenderer<V> {
             verticalScrollBar.setScrollMaximum(rows.size());
             verticalScrollBar.setViewSize(visibleRows);
             verticalScrollBar.setScrollPosition(viewTopRow);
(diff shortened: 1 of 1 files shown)


mabe02:
I'll put in some extra checks in case the table is moved to a different container in between drawings

avl42:
Not sure about those extra checks: Both verticalScrollBar and horizontalScrollBar are private, initialized in the constructor with no setters...  what other parent could they ever possibly have?

But then again, if there ever was a request for users to be able to supply their own scrollbars, then we at least won't need to worry about those parts anymore ;-)

ekoutanov:
It's conceivable that a component could be re-parented. For example, one could be rebuilding the complete GUI due to a user-requested layout change, which _may_ involve keeping the table but rebuilding the outer panels. (I'm thinking a text based email client, and the user wants to go from a vertical split to a horizontal split, or something along those lines.) The code in DefaultTableRenderer (after the fix) always explicitly re-assigns the parent, so I think it should be fine.

In any case, if the change is made to DefaultTableRenderer, it would make sense to consider all other components that feature scrollbars. TextBox does the same thing and it too, could be re-parented.

avl42:
right, didn't think of that, but right.

PR #271, opened 2016-10-05
AbstractScreen shouldn't pad CJK. BasicImage already did.

avl42:
I noticed that CJKScreenTest no longer showed any double-width characters.

The problem was CJK-padding at multiple layers:  At the bottom (BasicImage) it made sure that the character was padded, and that any double-width character immediately to the left of current position was removed.

Then, further up the call stack, AbstractScreen also tried to pad CJK, which meant it placed a space after it, which the BasicImage noticed to follow a CJK and thus removed the CJK.

This PR should fix it, and hopefully doesn't introduce regressions elsewhere.


mabe02:
Fingers crossed... :)

PR #273, opened 2016-10-05
Text graphics writer

avl42:
This is the new TextGraphicsWriter sitting on top of TextGraphics and rendering flowing rich text even with word wrapping.  The Test is a modified copy of the CJK-test to show off what it can do.

evaluate it, and if you like it, please add comments... (I'll be mostly offline for the rest of the week and didn't want it resting on my hard disk.)

If you pull this, it already contains the other two recent PRs.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/TerminalTextUtils.java
+++ b/src/main/java/com/googlecode/lanterna/TerminalTextUtils.java
@@ -23,7 +23,6 @@ import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 
-import com.googlecode.lanterna.TextColor.ANSI;
 import com.googlecode.lanterna.graphics.StyleSet;
 
 /**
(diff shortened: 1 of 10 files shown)


avl42:
Next step would be to integrate it into Label, but care needs to be taken:  Some labels are used to display "user-provided"(aka: "tainted") strings, and if such strings are crafted to contain certain sequences, they could "break" applications.  One way to address this would be to add a flag to the label for whether or not CSI sequences would be used.

any alternative ideas?

mabe02:
I agree in principal, just need some time to go through all the code...

jslater89:
Seems like a great solution to me, with lots of room for further rich text support down the road, too.

avl42:
Thanks for the feedback.  There are a couple of spots I intend to improve on, next time I get into a coding frenzy ;-)

I want the word/char/no-wrap behavior switchable, so I can pick any of a yet to be added enum WrapBehaviour with constants CLIP, CHAR, WORD.

I want the interpretation of escape codes switchable, and I want a two-char rendering of all control chars (that have no direct interpretation) as a caret(^) and the according letter, so e.g. the CR character \r would be rendered as "^M", and in case of TabBehaviour.IGNORE then a tab would be rendered as "^I". Ditto, if style-tagging is disabled, then \033 would render as "^[".

avl42:
one more WrapBehaviour:  SINGLE_LINE, which would cause even \n to be soft-rendered as "^J".

avl42:
mostly arrived:

 I can't break words yet that are longer than total width, though.

about Esc[21m: it worked on my gnome-terminal, and 21 (to undo bold 1) seems more logical ;-)
about Esc[m: split on an empty string still returns one element in the array.

new Test to really show off all the formatting features: style,tabs,escapes

have fun.

mabe02:
Merged!

PR #275, opened 2016-10-13
Mouseevent to Interactable mapping was flaky

### LINKED ISSUE #274: Mouseevents in odd positions crash lanterna-gui
On lanterna-discuss, Rajatt (thanks!) reported a bug that makes a lanterna-application with mouse-mode crash, if the user clicks on positions not covered by interactable components.

There are really two spots that need to be fixed, and I'll PR a fix, as soon as I have some issue-number to name the testcase accordingly.

avl42:
fixes #274


mabe02:
Thanks!

PR #280, opened 2016-11-04
Don't crash if terminal doesn't respond in time.

avl42:
This PR is more a request for discussion...

While playing with TerminalInputTest and mouse-capturing, it frequently happened that in the flood of mouse-events it missed the reply for the size query and the app crashed with IllegalStateException.  I don't know, whether the terminal just sent it too late, or missed sending the information at all, but it happened ways too often, while moving/dragging the mouse around.

This patch does three things:
 - wait up to 5 seconds instead of 2 seconds
 - if no reply found in waitForCursorPositionReport, return null instead of throwing up.
 - in ANSITerminal: deal with null by substituting dummy values: this could take further discussion. maybe we should try to just keep previous size, or just deliver "don't know" (aka: null) to users...


avl42:
further approaches could be to make it all asynchronous: don't wait, but register callbacks instead.

avl42:
This is now included in PR #282

PR #281, opened 2016-11-04
Remove item AbstractListBox/subclasses

jasonwjones:
Adds new capabilities to AbstractListBox: remove an item (with a given index), ability to remove an item by selecting it and pressing delete key, and ability to turn on/off the ability of the delete key (default is off)

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
@@ -150,7 +150,7 @@ public abstract class AbstractListBox<V, T extends AbstractListBox<V, T>> extend
                     return Result.HANDLED;
                     
                 case Delete:
-                	if (allowRemovesWithDelete) {
+                	if (allowRemovesWithDelete && selectedIndex > -1) {
                 		removeItem(getSelectedIndex());
                 		return Result.HANDLED;
                 	}
@@ -312,7 +312,7 @@ public abstract class AbstractListBox<V, T extends AbstractListBox<V, T>> extend
 	 *             equals/greater than the number of items in the list
 	 */
 	public synchronized T removeItem(int index) {
-		if (selectedIndex > -1 && !items.isEmpty()) {
+		if (index >= 0 && index < items.size()) {
 			items.remove(index);
 			if (selectedIndex >= items.size()) {
 				selectedIndex--;
--- a/src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
@@ -36,6 +36,7 @@ import java.util.List;
 public abstract class AbstractListBox<V, T extends AbstractListBox<V, T>> extends AbstractInteractableComponent<T> {
     private final List<V> items;
     private int selectedIndex;
+    private boolean allowRemovesWithDelete = false;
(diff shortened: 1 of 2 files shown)


src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
```diff
@@ -297,6 +303,44 @@ public synchronized V getSelectedItem() {
...
+	/**
+	 * Removes the item with the selected index from the list.
+	 * 
+	 * @param index the Index of the item to delete
+	 * @return Itself
+	 * @throws IndexOutOfBoundsException If the index is less than zero or
+	 *             equals/greater than the number of items in the list
+	 */
+	public synchronized T removeItem(int index) {
+		if (selectedIndex > -1 && !items.isEmpty()) {
```
jasonwjones: In retrospect, we can probably get by without the first check against selectedIndex. If `index is < 0` then the `IndexOutOfBounds` exception will just bubble up from the remove on the underlying list object. I think the original point of checking `selectedIndex` here was to ensure that the list box was the active component (although you wouldn't necessarily need the list box to be the active component to do a remove). So I'd say we can either replace `selectedIndex` with `index` or just remove that check altogether.

avl42:
Well, if someone hits the Del key in an empty listbox, I wouldn't want the
application to crash...

My vote would be in favor of both bounds checked for index and nothing
done, if outside.

jasonwjones:
I think maybe this adjustment makes things a bit better. A valid selectedIndex will be enforced when using the Delete key, but removeItem can be used irrespective of the current selection. Remove item checks that the index of the item to remove is valid with respect to the items list.

avl42:
perfect now :-) thanks!

we will still need to wait for @mabe02 to merge it.

mabe02:
Are you sure about this? Seems like expanding the API with an unusual edge case for me. It's possible to add a key-listener and drive this externally, no?

avl42:
I didn't notice for it until now... indeed, the AbstractInteractableComponent does have code to call onInput(...), but that can only return true (to pass the key to the component) or false (to bypass the component and treat the key as yet-unhandled.  It was designed as a filter, and that is not quite the same thing as a hook for custom key handlers.

jasonwjones:
The Delete key handling is a little bit of an edge case but given that it's off by default I thought that was a good compromise. What about list item removal in general though? I would have subclassed the list box but all of the fields are private so I thought this was better.

mabe02:
Yes, I think this is the real problem. There no way to remove a single item! I'll add a method for that, then you can extend and override handleKeyStroke.

mabe02:
How about that?

avl42:
Thanks for adding also what is necessary for the subclasses!

I still think, that removing an item with Delete key is "general purpose" enough that it shouldn't require sub-classing. Registering an inputFilter doesn't quite cut it, either. Maybe I just haven't found the "key-listener" feature yet that you mentioned, or we should add such ;-)

jasonwjones:
Looks mostly good but it looks like the methods all have a typo -- it should be removeItem(), not remoteItem(). Is there a way to easily add a key listener (to program in Delete handling) without having to subclass?

mabe02:
Ok, typo is corrected. Architecturally, I didn't go as far as making sure there were listeners for everything, I thought subclassing wouldn't be too complicated. There are some listeners spread around though, usually internal interfaces for some of the components. Maybe we can make something more generic and store it at the Component-level.

PR #282, opened 2016-11-07
Ansi terminal cursorquery+mousemode

avl42:
ANSITerminal
 - handle response timeout for cursor queries less strictly
 - limit mouse capture mode to private mode - reset it on exitPrivateMode() because the need to do it explicitly is quite inconvenient for applications.


avl42:
included in # 282

PR #288, opened 2016-11-16
Fix bug on LinearLayout with empty containers

nicoulaj:
This fixes a small bug in LinearLayout: if the container is empty, it creates a TerminalSize with -1, which crashes.


mabe02:
Looks good, thanks!

PR #291, opened 2016-12-04
Two small fixes

rhelmot:
- when working with a local copy of lanterna things were breaking because the BundleLocator wasn't searching for the resources in src/main/resources like [LanternaThemes is](https://github.com/mabe02/lanterna/blob/master/src/main/java/com/googlecode/lanterna/bundle/LanternaThemes.java#L112).
- The logic for deciding whether to draw a vertical scrollbar in the edge case that the text is exactly the same height as the view area but a horizontal scrollbar has been drawn seemed pretty broken, so I fixed it, I think?

### DIFF
--- a/src/main/java/com/googlecode/lanterna/bundle/BundleLocator.java
+++ b/src/main/java/com/googlecode/lanterna/bundle/BundleLocator.java
@@ -20,6 +20,7 @@ package com.googlecode.lanterna.bundle;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.FileInputStream;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.net.URLConnection;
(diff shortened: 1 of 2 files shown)


avl42:
The scrollbar-fix looks good to me.

I'm not feeling well about picking files from local file system as a fallback for files not found in the jar, if this feature is retained in production.

I could imagine, that users could misuse it to get custom code executed by merely asking for an exotic language and having placed well-crafted files into a src/... directory below the current one. 

Maybe we could find a way to limit this to "development environments" e.g. by checking a property.

PS: Themes may specify Renderer classes, so this isn't just purely theoretical.  Yes, I'm paranoid, but that doesn't mean **they** aren't after me ;-)

mabe02:
Agreed, we probably don't want loading from the local file system with any kind of path. We could add some kind of development mode through system properties, like com.googlecode.lanterna.developmentMode and using Boolean.getBoolean(..) to check it

avl42:
A recent change about adding a default theme directly into the code might solve some of the motivation for the ressource-part of this PR.

But maybe the scrollbar fix for TextBox could still be cherry-picked from it.

mabe02:
Yeah, let's do that (once I figure out how to cherry-pick on github...)

PR #296, opened 2017-03-03
Only check visible windows for invalidness

thumbnail:
Currently the `WindowTextGUI` only draws visible windows (since #201). Unfortunately this introduced a bug which caused lanterna redrawing until all windows are visible. This PR checks whether a window is invalid **and** visible, which should prevent this from happening.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
@@ -1,6 +1,6 @@
 /*
  * This file is part of lanterna (http://code.google.com/p/lanterna/).
- * 
+ *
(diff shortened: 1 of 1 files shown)


avl42:
I'm a bit surprised how invisible windows end up invalidated in the first place, but it certainly may happen and it shouldn't lead to an always busy process.

thumbnail:
Well, the dirty-tracking isn't taking the invisibility into account is it? I think that shouldn't be the case either since it'll complicate things and the component is technically dirty when  component changes. Whether it is being drawn or not.

Anyway, I came across this problem when i set the visibility of a window on `false` in the constructor.

avl42:
You're right.  The fix is simple and looks good, but still we need to wait for @mabe02 to merge it.

mabe02:
Thanks!

PR #297, opened 2017-03-04
update method name

cvogt:
thx a lot for lanterna, great stuff :)


avl42:
That seems to be one of a few documents that were originally written for version 2.x (if not 1.x), and weren't yet updated for 3 .

Another change is, that a class named Key no longer exists - instead there are now KeyStroke and KeyType classes.

Thanks for making a start, PR'ing some of these changes.

PR #303, opened 2017-04-09
Allow focus to pass back to background pane

brobo:
This commit only affects instances of `MultiWindowTextGUI`.
In the case that the background pane has interactable components, it may be desirable to allow focus to pass to the background pane. This is currently fully supported so long as no windows are opened, but as soon as a window is opened it is no longer possible to return focus to the background pane.

`setActiveWindow(null)` now allows the background pane to have focus, and if no suitable window can be found to replace a removed active window, then focus returns to the background pane.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
@@ -379,15 +379,18 @@ public class MultiWindowTextGUI extends AbstractTextGUI implements WindowBasedTe
         }
         window.setTextGUI(null);
         windowManager.onRemoved(this, window, windows);
-        if(activeWindow == window) {
+        changeWindow: if(activeWindow == window) {
             //Go backward in reverse and find the first suitable window
             for(int index = windows.size() - 1; index >= 0; index--) {
                 Window candidate = windows.get(index);
                 if(!candidate.getHints().contains(Window.Hint.NO_FOCUS)) {
                     setActiveWindow(candidate);
-                    break;
+                    break changeWindow;
                 }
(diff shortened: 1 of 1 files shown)


mabe02:
Looks good, thanks!

src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
```diff
@@ -471,11 +474,11 @@ public synchronized WindowBasedTextGUI moveToTop(Window window) {
...
-        if(windows.isEmpty() || windows.size() == 1 || activeWindow.getHints().contains(Window.Hint.MODAL)) {
+        if(windows.isEmpty() || windows.size() == 1 || (activeWindow != null && activeWindow.getHints().contains(Window.Hint.MODAL))) {
             return this;
         }
         Window originalActiveWindow = activeWindow;
-        Window nextWindow = getNextWindow(reverse, originalActiveWindow);
+        Window nextWindow = activeWindow == null  ? windows.get(0) : getNextWindow(reverse, activeWindow);
         while(nextWindow.getHints().contains(Window.Hint.NO_FOCUS)) {
             nextWindow = getNextWindow(reverse, nextWindow);
             if(nextWindow == originalActiveWindow) {
```
avl42: I suspect from just looking at it, that cycling away from <null>, if all Windows are not focus-able, might result in an infinite loop, because getNextWindow() only cycles among the (not null) windows, but originalActiveWindow is null here....

mabe02: You are probably right, I'll add something to address this

PR #306, opened 2017-04-14
UnixTerminal should now close properly

### LINKED ISSUE #305: Closing UnixTerminal doesn't reset keyEcho and specialCharacters
Consider the following example:
```
package bugtest;

import com.googlecode.lanterna.terminal.DefaultTerminalFactory;
import com.googlecode.lanterna.terminal.Terminal;
import java.io.BufferedReader;
import java.io.IOException;
import java.io.InputStreamReader;

public class Bugtest {

    /**
     * @param args the command line arguments
     */
    public static void main(String[] args) throws IOException, InterruptedException{
        Terminal terminal = new DefaultTerminalFactory().createTerminal();
        System.out.println("Class: " + terminal.getClass());
        
        terminal.close();
      
        BufferedReader br = null;
        try {
            br = new BufferedReader(new InputStreamReader(System.in));

            while (true) {

                System.out.print("Enter something: ");
                String input = br.readLine();

                if ("q".equals(input)) {
                    System.out.println("Exit!");
                    System.exit(0);
                }

                System.out.println("input : " + input);
                System.out.println("-----------\n");
            }

        } catch (IOException e) {
            e.printStackTrace();
        } finally {
            if (br != null) {
                try {
                    br.close();
                } catch (IOException e) {
                    e.printStackTrace();
                }
            }
        }

    }
    
}
```
After closing the terminal, characters typed don't appear on the screen although the buffered reader can print them, and there is worse: you can't kill the process since control characters aren't working (only CTRL+Z is working, so you can put it to background, do `fg`, and control characters will work again (as well as echo)).
I tested it with xterm and lxterminal, I don't know if it happens with a Swing or another terminal.
I think I found a way to patch UnixTerminal so if it works I will make a pull request.

rmelotte:
UnixTerminal should now close properly: restoring keyEcho, keyStrokeSignals and terminalSettings.
**Please don't merge this PR without reviewing it, as I don't know if I did it properly.**
(I tested it and it works against my needs though)
Should fix #305

### DIFF
--- a/src/main/java/com/googlecode/lanterna/terminal/ansi/UnixTerminal.java
+++ b/src/main/java/com/googlecode/lanterna/terminal/ansi/UnixTerminal.java
@@ -165,4 +165,12 @@ public class UnixTerminal extends UnixLikeTTYTerminal {
         
         return super.findTerminalSize();
     }
+    
+    @Override
+    public void close() throws IOException{
+        super.close();
+        restoreTerminalSettings();
+        keyEchoEnabled(true);
+        keyStrokeSignalsEnabled(true);
+    }
 }


mabe02:
Okay, I'm not sure if this is in the right place. I'll try it out. The difficulty here is knowing what state the terminal is in when you start the application and trying to restore the state to that when it exits. You'd assume we can just put it back to "normal" mode (key echo enabled, canonical mode on, no control character traps, etc) but this might not be the case.

mabe02:
Okay, I see what's happening now. Lanterna currently restores the terminal when the JVM exits. I'll see if I can get it to do it both both close() and JVM exit.

mabe02:
I've implemented and pushed this now, please see 713514bc

PR #308, opened 2017-05-17
Bugfix: resizing terminals not working in some layout managers

jbergenthal2:
Bugfix for resizing Terminals in layout managers that ask for preferred size, like BorderLayout.
If we put a Terminal inside the centre region of a Border layout, it stretches horizontally and vertically when resized.
However, the Terminal size (character size) remains the same as before resizing.
This happens because before painting, the layout manager asks for the Terminal preferred size (which has the old size value).
What needs to be done is to update the terminal size as soon as the Terminal is resized (in the setBounds() call).
I hope this description makes sense.


jbergenthal2:
I made a few more changes, will file another pull request

PR #309, opened 2017-05-24
DirectoryDialog and simple drop-down menus

fracpete:
* Added support for selecting directories, similar to the `FileDialog` class, using the `DirectoryDialog` class. Builder class is also available: `DirectoryDialogBuilder`
* Added support for very simple drop-down menus (menubar, menu, menuitem)
* Added example code and documentation for both features


mabe02:
Okay, I need to spend more time reviewing this. Probably I'll branch off current master to a release/3.0 and let this go into the 3.1 release.

Issue #327, opened 2017-08-24
Can not alter TerminalEmulatorDeviceConfiguration


klaushauschild1984:
When I use the default of TerminalEmulatorDeviceConfiguration then it is not possible to alter its properties, I have to instantiate a new one with all the default values copied into my code only for changing one thing.
Please extend it with the builder pattern so that the properties can be altert after creation.

mabe02:
Yes, should be done

klaushauschild1984:
I opend a PR.

PR #328, opened 2017-08-27
Can not alter TerminalEmulatorDeviceConfiguration #327

klaushauschild1984:
- add withBlinkLengthInMilliSeconds
- change withLineBufferScrollbackSize: do not copy the TerminalEmulatorDeviceConfiguration, set the private fiels
- add withCursorStyle
- add withCursorColor
- add withCursorBlinking
- add withClipboardAvailable

### DIFF
--- a/src/main/java/com/googlecode/lanterna/terminal/swing/TerminalEmulatorDeviceConfiguration.java
+++ b/src/main/java/com/googlecode/lanterna/terminal/swing/TerminalEmulatorDeviceConfiguration.java
@@ -157,7 +157,8 @@ public class TerminalEmulatorDeviceConfiguration {
                     blinkLengthInMilliSeconds,
                     cursorStyle,
                     cursorColor,
-                    cursorBlinking);
+                    cursorBlinking,
+                    clipboardAvailable);
         }
     }
 
@@ -175,7 +176,8 @@ public class TerminalEmulatorDeviceConfiguration {
                     blinkLengthInMilliSeconds,
                     cursorStyle,
                     cursorColor,
-                    cursorBlinking);
+                    cursorBlinking,
+                    clipboardAvailable);
         }
     }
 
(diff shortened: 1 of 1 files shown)


RAnders00:
Have you explicitly considered the implications of making that class mutable? How do the classes that use this object behave when properties of this class are changed?

Also, it is generally convention that `withXXX()` methods return a copy of a immutable object with that property changed. If you really wanted to make the class mutable, the method signature should at least be changed to `setXXX()` methods with a return type of `void`.

avl42:
Fwiw (not much) I agree with Ruben.  If we really want it mutable, it
should have "set"ters rather than "with"ers :-)

Anyway, creating parameter sets for terminalemulators shouldn't be expected
in tight loops so shouldn't be performance critical.
Having some fields copied into an immutable class shouldn't hurt - as long
as they are copied within lanterna code, and not left to the user to copy.

A while ago I tried to revise that part, but I got carried away and it
became ways worse than status quo w.r.t. complexity. I'm looking forward to
what this discussion will lead to.
​I certainly agree to OP that there is indeed some potential for
improvement.

klaushauschild1984:
No, it was not my intension to change TerminalEmulatorDeviceConfiguration from a immutable object to a mutable one. I was just on the shamefull road of bad API design.
A real Build seam to over engineered so I go withe the every-time-a-new-object idea.

avl42:
I have a minor "complaint" about style, though: I'd rather see some (even redundant) "this."-prefixes each time the current object's attribute is used versus where the specified new value is used (without "this." of course).

klaushauschild1984:
@avl42 can you point out where your complaints are? Currently I see them not.

avl42:
I will, but I'm on phone right now...

Am 28.08.2017 18:17 schrieb "Klaus Hauschild" <notifications@github.com>:

klaushauschild1984:
Now I got it. You WANT the redundant this's for visibility.

avl42:
Yes, that's what I meant. Thanks!

I hope, @mabe02 also appreciates it.

mabe02:
Thanks!

PR #331, opened 2017-09-02
fix package name in documentation

ljosa:
DefaultTerminalFactory is in com.googlecode.lanterna.terminal (https://github.com/mabe02/lanterna/blob/master/src/main/java/com/googlecode/lanterna/terminal/DefaultTerminalFactory.java#L19).


mabe02:
Thanks!

Issue #332, opened 2017-09-02
readInput on Terminal from TelnetTerminalServer returns immediate EOF


ljosa:
Thanks for Lanterna! I'm trying to use it as a telnet server, and I'm running into an issue that I think may be a problem with Lanterna.

**Version**: `[com.googlecode.lanterna/lanterna "3.0.0"]` (jar from Maven)

**Code that demonstrates the problem:**

```java
import com.googlecode.lanterna.terminal.*;
import com.googlecode.lanterna.terminal.ansi.*;

class DemoTelnet {
     public static void main(String args[]) throws java.io.IOException {
          TelnetTerminalServer server = new TelnetTerminalServer(8050);
          Terminal t = server.acceptConnection();
          System.out.println(t.readInput().toString());
          t.close();
     }
}
```

**Directions to reproduce:**

1. Start server with the code above.

2. `telnet localhost 8050`

**Expected behavior:** wait for the telnet user to press a key, then close the connection.

**Observed behavior**: closes the connection immediately, without waiting. The program prints `KeyStroke{keytype=EOF}`.

**Notes**

The following code, which is similar but uses `DefaultTerminalFactory`, works as expected in a terminal on macOS (where the factory returns a UnixTerminal).

```java
import com.googlecode.lanterna.terminal.*;

class DemoText {
     public static void main(String args[]) throws java.io.IOException {
          Terminal t = new DefaultTerminalFactory().createTerminal();
          System.out.println(t.readInput().toString());
          t.close();
     }
}
```

avl42:
This is also reproducible on Linux, and the problem is, that
TelnetClientIACFilterer (nested within TelnetTerminal) doesn't do a proper
blocking read() - instead, if no input is available then it just returns -1
rather than wait for input.

Btw., this problem doesn't show in the TelnetTerminalTest, because that
test only does non-blocking reads.

Good catch!

PR #333, opened 2017-09-03
Make read() block in TelnetTerminal

ljosa:
I took a stab at fixing #332 ("readInput on Terminal from TelnetTerminalServer returns immediate EOF"). This also adds a demonstration of blocking input to TelnetTerminalTest.

Thanks for locating the code that needed to change, @avl42!

Please review carefully, as there is lots that I don't understand about the TELNET protocol and the Lanterna codebase.


avl42:
the fix looks perfect to me!

mabe02:
Good catch!

Issue #353, opened 2018-03-07
UnsupportedOperationException on MessageDialogBuilder#setExtraWindowHints


andysenn:
Hello,

First of all, I love Lanterna. Thank you so much for your hard work on this project.

When I call MessageDialogBuilder#setExtraWindowHints(...), it throws an UnsupportedOperationException. It looks like this is because the Iterator<T> from the underlying Set<T> implementation for hints does not support the #remove() operation.

Here is the full stack trace:

```
Exception in thread "main" java.lang.UnsupportedOperationException
	at java.util.Collections$1.remove(Collections.java:4684)
	at java.util.AbstractCollection.clear(AbstractCollection.java:436)
	at com.googlecode.lanterna.gui2.dialogs.MessageDialogBuilder.setExtraWindowHints(MessageDialogBuilder.java:94)
	at com.jamfsoftware.tools.jss.initializer.ui.cli.CLIApplication.showMessageBox(CLIApplication.kt:266)
	at com.jamfsoftware.tools.jss.initializer.ui.cli.CLIApplication.validate(CLIApplication.kt:249)
	at com.jamfsoftware.tools.jss.initializer.ui.cli.CLIApplication.execute(CLIApplication.kt:257)
	at com.jamfsoftware.tools.jss.initializer.ui.cli.CLIApplication.access$execute(CLIApplication.kt:20)
	at com.jamfsoftware.tools.jss.initializer.ui.cli.CLIApplication$start$$inlined$apply$lambda$14.run(CLIApplication.kt:103)
	at com.googlecode.lanterna.gui2.Button$1.onTriggered(Button.java:74)
	at com.googlecode.lanterna.gui2.Button.triggerActions(Button.java:101)
	at com.googlecode.lanterna.gui2.Button.handleKeyStroke(Button.java:93)
	at com.googlecode.lanterna.gui2.AbstractInteractableComponent.handleInput(AbstractInteractableComponent.java:142)
	at com.googlecode.lanterna.gui2.AbstractBasePane.doHandleInput(AbstractBasePane.java:126)
	at com.googlecode.lanterna.gui2.AbstractBasePane.handleInput(AbstractBasePane.java:97)
	at com.googlecode.lanterna.gui2.AbstractWindow.handleInput(AbstractWindow.java:133)
	at com.googlecode.lanterna.gui2.MultiWindowTextGUI.handleInput(MultiWindowTextGUI.java:333)
[...]

PR #354, opened 2018-03-08
Fix and Test for Issue #353

avl42:
This fixes an exception thrown for trying to supply Window.HINTs to the MessageDialogBuilder.

Also, change MessageDialogBuilder.setExtraWindowHints to accept any Collection<Window.HINT>, as there is really no need to confine to a Set.


mabe02:
Looks good, thanks!

PR #339, opened 2017-11-13
Allow empty action list in ActionListDialog

asoltysik:
Issue #338 
The condition to require a non-empty action list seemed rather arbitrary to me, so I'll leave this PR for your consideration.

Issue #338, opened 2017-11-13
ActionListDialog needs at least one action


asoltysik:
What's the reasoning behind the behaviour mentioned in the title? Is there needed or is it just a design decision? I would like to have an empty `ActionListDialog` with just cancel button in case I have nothing to show.

avl42:
Good question.  For testing, modify the ActionListDialog to get rid of that check, and see if the ActionListBox used in the Dialog is specifically unhappy about absence of items.  The Dialog might just protect the Box.

If it all works fine then please consider a PR to remove the check.

asoltysik:
Done, #339.

mabe02:
I thought there will be some oddities with focus where the component has focus but there is no visual indication of it?

mabe02:
Nevermind, seems to work. Just weird to have a dialog without any items.

PR #366, opened 2018-05-19
allow empty action list in ActionListDialog

avl42:
This is @asoltysik's PR #339 rebased to release/3.0

> Issue #338
> The condition to require a non-empty action list seemed rather arbitrary to me, so I'll leave this PR for your consideration.


mabe02:
I think it behaves a bit weird with no elements when it's focused, I think that's why I put it in

mabe02:
Nevermind, seems to work ok

Issue #351, opened 2018-02-26
high-DPI detection of AWTTerminalFontConfiguration is broken on my system


jrb0001:
I have two monitors at 1920x1080 each, but Java reports them as `[3840, 1080]` and `[1920, 1080]` which triggers the extremely high font size. Wouldn't it be a better idea to use `Toolkit.getDefaultToolkit().getScreenResolution()` and then maybe 160 as threshold? If that's not possible, maybe the combination of `width > 2000 && height > 1400`?

avl42:
I'm not sure I understand it correctly, but Lanterna might have a hard time working around Java problems.  If you have a particular workaround in mind, where Lanterna could indeed solve it, then a PR would be welcome, provided it doesn't introduce problems for "normal" screen setups.

jrb0001:
Changing something without knowing why it was done in the current way is a bit meh. Is there any particular reason why it was done the current way or was it simply the first idea?

avl42:
I can only guess: maybe historical experience with queried screen
resolutions were bad enough to rather rely on other heuristics.

Maybe nowadays DPI values are better quality nowadays and beat other
heuristics.  They're probably better on mobile devices (phones, tablets)
where the driver "knows" the physical screen, and still mostly guesswork
where the driver doesn't know the physical screen.

So much for my speculation about status-quo ;-)

avl42:
One of my machines' graphics port is connected to a VGA-splitter that
drives both a 22" monitor and a 40" TV-screen.  What would be a correct DPI
value?

mabe02:
Relevant code (from `AWTTerminalFontConfiguration`):
```
        // Assume the first GraphicsDevice is the primary screen (this isn't always correct but what to do?)
        // Warning, there could be printers coming back here according to JavaDoc! Hopefully Java is reasonable and
        // passes them in after the real monitor(s).
        if (gs.length > 0) {
            int primaryMonitorWidth = gs[0].getDisplayMode().getWidth();

            // If the width is wider than Full HD (1080p, or 1920x1080), then assume it's high-DPI
            if (primaryMonitorWidth > 2000) {
                CHOSEN_FONT_SIZE = 28;
            }
        }

        // If no size was picked, default to 14
        if(CHOSEN_FONT_SIZE == null) {
            CHOSEN_FONT_SIZE = 14;
        }
```
[...]

PR #369, opened 2018-05-20
Rework high-DPI detection to work in multi-monitor setups on linux (#351)

jrb0001:


src/main/java/com/googlecode/lanterna/terminal/swing/AWTTerminalFontConfiguration.java
```diff
@@ -70,61 +76,54 @@
...
-                CHOSEN_FONT_SIZE = 28;
+    private static int getFontSize() {
+        if (Toolkit.getDefaultToolkit().getScreenResolution() >= 110) {
+            // Rely on DPI if it is a high value.
+            return Toolkit.getDefaultToolkit().getScreenResolution() / 7 + 1;
+        } else {
+            // Otherwise try to guess it from the monitor size:
+            // If the width is wider than Full HD (1080p, or 1920x1080), then assume it's high-DPI.
+            GraphicsEnvironment ge = GraphicsEnvironment.getLocalGraphicsEnvironment();
+            if (ge.getMaximumWindowBounds().getWidth() > 4000) {
```
avl42: Since we're not using screen-resolutions but MaximumWindowBounds, I think 4000 is an unfortunate value.  Someone with a 4096 monitor then might see different font-sizes in lanterna apps based on how wide his sidebar is. E.g.: 4096 with a 96 pixel wide sidebar will make a smaller font than with an even-slightly-more-than-96 pixel wide sidebar.

If there isn't a convenient "the widest one of all the monitors" method, I'd propose 3840 as a somewhat "sweeter" value.

avl42: Here's a chart of standard monitor sizes:  https://en.wikipedia.org/wiki/Display_resolution#/media/File:Vector_Video_Standards8.svg

jrb0001: Yes, I completely forgot that somebody might have a sidebar. I have reduced it to >3840 and >1920 as suggested.

jrb0001: I thought about it again and I think that 2k shouldn't be high-DPI yet. So a threshold of >4096 and >2048 should be better? That would give 512px for the sidebar for UWHD/WQHD/WXGA/QSXGA.

src/main/java/com/googlecode/lanterna/terminal/swing/AWTTerminalFontConfiguration.java
```diff
@@ -70,61 +76,54 @@
...
-            int primaryMonitorWidth = gs[0].getDisplayMode().getWidth();
-
-            // If the width is wider than Full HD (1080p, or 1920x1080), then assume it's high-DPI
-            if (primaryMonitorWidth > 2000) {
-                CHOSEN_FONT_SIZE = 28;
+    private static int getFontSize() {
+        if (Toolkit.getDefaultToolkit().getScreenResolution() >= 110) {
+            // Rely on DPI if it is a high value.
+            return Toolkit.getDefaultToolkit().getScreenResolution() / 7 + 1;
+        } else {
```
mabe02: Not sure the `else` makes sense anymore...?
So if the DPI is lower than 110, we're still doing the previous heuristics based on screen resolution? I really don't remember how reliable the DPI detection in Java is, would it be returning something below 110 if it doesn't know and that's why we should keep the below logic?

jrb0001: The default value is somewhere around 90/96 on most older systems afaik. I think it's safe to keep the heuristics for a while as high DPI support by some OSes is still a bit new.

mabe02:
Let's retarget this for release/3.0 and include it in the next build

jrb0001:
Retargeted for release/3.0.

PR #377, opened 2018-06-17
Fix AbstractComposite.nextFocus for disabled components

### LINKED ISSUE #374: Disabled components in a border mess up focus traversal
When you have a disabled component (setEnabled(false)) in a border (addComponent(other.withBorder()) and it gets hit by find-next-thing-focus traversal it returns it as focusable, eventually it gets checked for being enabled and discarded and focus does not traverse anywhere making TAB effectivelly broken.

(I am just writing this down here so that I do not forget, test-case, and PR with a fix coming soon.)

honzasterba:
- composite should only offer its component for focus if given component is enabled
- a disabled component in a border breaks forcus traversal

fixes #374


honzasterba:
good point, yes it should!

mabe02:
Looks reasonable, thank you!

PR #379, opened 2018-06-21
Fix #378

### LINKED ISSUE #378: [FileDialog] When an absolute file name is entered in fileBox, the file should not be relative to directory in directoryListBox.
When I open a FileDialog, it's sometimes faster to type an absolute path in the fileBox rather than navigating to the directory in the directoryListBox.

However the converted path is **always** relative to the directory.

I believe the path should be considered absolute whenever it starts with '/' (in Linux).

Example :
capture

I expect the resulting path to be "/dev/sda1" however I get "/home/engineer/Documents/GitHub/fatChains/dev/sda1"

lolo101:
When the path entered in fileBox is absolute, the directory from directoryListBox is ignored.


mabe02:
Thank you!

Issue #387, opened 2018-07-18
Table and PgUp & PgDn support


MartyIX:
Hi,

I have noticed that [`Table`](https://github.com/mabe02/lanterna/blob/master/src/main/java/com/googlecode/lanterna/gui2/table/Table.java) does not support `PgDn` and `PgUp` keys. It would be nice to support them. 

Having a quick glimpse at the code, it does not seem so hard to add the functionality and I may be able to do provide a PR.

Best regards,
Martin

mabe02:
Contributions are much welcome!

PR #389, opened 2018-07-18
Table component - add support for PgUp and PgDn

MartyIX:
Hi,

This is a proposed PR for #387 issue. What do you think?

Regards,
Martin

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/table/Table.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/table/Table.java
@@ -369,6 +369,19 @@ public class Table<V> extends AbstractInteractableComponent<Table<V>> {
                     return Result.MOVE_FOCUS_DOWN;
                 }
                 break;
+            case PageUp:
+                if(visibleRows > 0 && selectedRow > 0) {
+                    selectedRow -= Math.min(getVisibleRows(), selectedRow);
+                    viewTopRow = selectedRow;
+                }
+                break;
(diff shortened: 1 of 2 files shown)


mabe02:
Looks good to me, thanks!

MartyIX:
@mabe02 There is one thing that is not solved. When client code does not set `setVisibleRows(num)`, then PgUp & PgDn do not work. It's because I'm not sure how to obtain the number of visible rows in that case.  

I must say that `visibleRows` field is a bit confusing to me. I would expect that it always contain value that corresponds witht what user can see.

mabe02:
If you don't set this value, the table will request to be large enough to display all the rows. If the terminal window isn't large enough to support this, the size will ultimately be decided based on how many rows does fit.

mabe02:
Okay, a couple of hours later I've re-written half of the table and while it solves the problems above it's probably got a million other bugs. I've put it on master for now. For 3.0, I think we're going to have to admit that PageUp and PageDown won't work for tables that are not vertical size restricted. The scrolling doesn't work properly anyway so there wouldn't be much point in trying to fix the paging.

MartyIX:
@mabe02 
> Okay, a couple of hours later I've re-written half of the table and while it solves the problems above it's probably got a million other bugs. I've put it on master for now. 

Great 👍 

> The scrolling doesn't work properly anyway so there wouldn't be much point in trying to fix the paging.

How do you mean?

PR #401, opened 2018-10-12
Implement TableModel.clear() method

MartyIX:


MartyIX:
@mabe02 Do you have any comments on this?

PR #406, opened 2018-11-25
Maven Surefire Plugin

BrunoEberhard:
could not build without it.


mabe02:
Never had any problems with it. 3.0.0-M1 is _very_ recent, I'm a bit hesitant to enforce it. Did you try building with Maven from command line?

BrunoEberhard:
Tried again from command line. Works. Sorry for that one.

Issue #417, opened 2019-05-28
waitUntilClosed don't return or throw exception when socket is closed.


realkcn:
I write a telnet server like the example [com/googlecode/lanterna/gui2/GUIOverTelnet.java](https://github.com/mabe02/lanterna/blob/master/src/test/java/com/googlecode/lanterna/gui2/GUIOverTelnet.java).
When the client use ctrl+] to force closing telnet socket, the thread don't exit.
Seem it is infinite loop in function waitForWindowToClose and can't detect the socket is closed.

Call Stack below:
"Thread-0@590" prio=5 tid=0xd nid=NA sleeping
  java.lang.Thread.State: TIMED_WAITING
	  at java.lang.Thread.sleep(Thread.java:-1)
	  at com.googlecode.lanterna.gui2.MultiWindowTextGUI.waitForWindowToClose(MultiWindowTextGUI.java:421)
	  at com.googlecode.lanterna.gui2.AbstractWindow.waitUntilClosed(AbstractWindow.java:269)
	  at com.googlecode.lanterna.gui2.MultiWindowTextGUI.addWindowAndWait(MultiWindowTextGUI.java:373)
	  at cn.kcn.term.server.TestMain.runGUI(TestMain.java:157)
	  at cn.kcn.term.server.TestMain.access$000(TestMain.java:35)
	  at cn.kcn.term.server.TestMain$1.run(TestMain.java:47)

avl42:
My own single-stepping test with GUIOverTelnet makes it look like a ServerSocket just simply doesn't let us know that the client is gone, if the client types Ctrl-] and "close" in telnet. 

I got as far as stepping through AbstractPlainSocketImpl.available() and all it reports is "no data at the moment" - specifically it hasn't even noticed that the socket is broken.

It seems to me, at least at this time, that Lanterna cannot do much about it.

realkcn:
> My own single-stepping test with GUIOverTelnet makes it look like a ServerSocket just simply doesn't let us know that the client is gone, if the client types Ctrl-] and "close" in telnet.
> 
> I got as far as stepping through AbstractPlainSocketImpl.available() and all it reports is "no data at the moment" - specifically it hasn't even noticed that the socket is broken.
> 
> It seems to me, at least at this time, that Lanterna cannot do much about it.

It need some notify to break the windows event loop.
TelnetTerminal.available() method should throw EOFException when socket is closed. waitForWindowToClose method will catch the EOFException and quit the loop.

Is it possible?

[...]

PR #428, opened 2019-08-17
Issue #417: Detect client abruptly closing the socket

mabe02:
This is an issue with how .available() is implemented for sockets, it's actually not really reliable and you are expected to call .read() instead. This adds an adapter class that transforms the .available() calls to .read()


avl42:
Oh, and I think you can close this one.

Issue #422, opened 2019-06-10
Please ship WindowsTerminal along with rest of the binary classes


unix-junkie:
Currently, the version of the library available from the _Maven Central_ lacks `WindowsTerminal` class.

Please ship `WindowsTerminal` plus native code for win32/win64 along with the rest of the library.

Additionally, _lanterna_ can have an optional dependency on the _native-platform_ library which already ships the native primitives for all mainstream platforms. `WindowsTerminal` could be re-implemented via those primitives -- see [`WindowsConsoleFunctions.java`](https://github.com/adammurdoch/native-platform/blob/master/src/main/java/net/rubygrapefruit/platform/internal/jni/WindowsConsoleFunctions.java).

mabe02:
Unfortunately, the native bindings library was more of an experiment and was never 100% functional. As long as it's experimental, I don't think it would be a good idea to distribute it to Maven central.

avl42:
if people on windows OS would test & PR this module into shape, then things
might be different...

mabe02 <notifications@github.com> schrieb am So., 18. Aug. 2019, 15:22:

rednoah:
@mabe02 I had a crack at it over the weekend and got it working, after many many hours of debugging.

Since `lanterna` is already doing `Class.forName(...)` this can be added as 3rd party library, but you're welcome to merge it into your code base here:
https://github.com/rednoah/lanterna-wincon

They key was figuring out that on Windows the `InputStream.available()` method of `System.in` is basically lying, and so our input loop breaks in strange unexpected ways when it randomly starts blocking in code paths that shouldn't be blocking.

mabe02:
Interesting! Did you write this from scratch or did you base it off the experimental work we've done in native-integration?

rednoah:
I started out with your code base to debug the issue. The new code just about rewriting things to use `ReadConsoleInput` and `GetNumberOfConsoleInputEvents` instead of `System.in.read()` and `System.in.available()` (which just doesn't work) and to get `resize` events. Using `WriteConsole` for output helped with screen flickering and allows for printing Emoji.

rednoah:
`jna-platform` now includes the added native bindings as well:
https://github.com/java-native-access/jna/commit/b22f85a4e550fd3862855b2575bfb45105572638

mabe02:
[...]

PR #429, opened 2019-08-19
Modularize for JPMS

jrb0001:
The PR has two commits and I recommend reviewing them individually:
- Packaging changes to create a multi-release jar file that works on Java 6+ and is a proper JPMS module (Java 9+). This bumps the minimum javac version to 9 but the resulting jar should work on Java 6 (untested).
- Remove the runtime dependency on the `java.desktop` module. This allows producing smaller runtime images if you don't need the terminal emulator parts nor the `TextColor.getColor()` method.

I tested it on Java 11 with a modularized version of the `OutputString` example both with and without the `java.desktop` module.

Is the package `com.googlecode.lanterna.bundle` meant to be used directly by the application/other libraries? If not, then its `exports` should be dropped.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -34,6 +34,13 @@
             <version>4.12</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.sun.xml.ws</groupId>
+            <artifactId>jaxws-ri</artifactId>
+            <version>2.3.2</version>
+            <type>pom</type>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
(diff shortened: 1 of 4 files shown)


jrb0001:
Please note that this breaks `WindowsTerminal` if lanterna is used on the modulepath instead of the classpath. One way to solve that would be #422 and the alternative would be to move it to its own package and then use SPI to load it.

SPI sound like the better approach to me because that would also allow moving the terminal emulator part into its own module with proper runtime dependencies to `java.desktop`.

mabe02:
I don't care too much about breaking the experimental, unfinished native bindings

Issue #430, opened 2019-09-25
First "F3" not working.


bleehome:
We use key "KeyType.F3" for shortcut of searching action, and the first press of F3 has no effect.
![image](https://user-images.githubusercontent.com/1271744/65575905-74938900-dfa3-11e9-8407-9d2432d898f5.png)
![image](https://user-images.githubusercontent.com/1271744/65575938-88d78600-dfa3-11e9-9ca2-7af4b5cfe28e.png)

F3 is used for cursor position reporting?

bleehome:
I set a position to lastReportedCursorPosition by :

![image](https://user-images.githubusercontent.com/1271744/65577764-0ea90080-dfa7-11e9-82c8-14668daadd85.png)

but it is ugly.

avl42:
You're right, that it is unfortunate, that designers of terminals came up with the questionable "idea" to send the same sequence for cursor-position queries as for user typing a specific key.

Lanterna tries hard not to accidentally misinterpret a cursor-report for a spurious F3, but there is some room for error the other way, if the F3 key is already hit, before the application shows up on the screen.

avl42:
Something could be changed to at least get un"shift"ed,un"control"ed,un"alt"ed F3, which (at least on some terminals I checked) is distinguishable from a cursor-report.   It would mean adding a flag to KeyStroke, and set it for F3, depending on whether it came from "^[OR" or from "^[[1;1R".   Maybe, even "^[[1;1R" is a special case that doesn't ever come from terminals, thus could be special-cased to not give an F3, and then we could restrict the reverse-conversion (second screenshot).

@mabe02 ?

mabe02:
Yikes, I had even forgotten about this... So @bleehome is this consistently never working the first time? Would it be possible for you to put together a small sample test to reproduce this?

avl42:
The problem here is that his F3 keypress comes before the requested
response for terminal cursor position - for determining terminal size, and
because F3-escape codes may look like terminal position responses, lanterna
tends to map apparent F3 keys back to position reports, if one is expected.

mabe02 <notifications@github.com> schrieb am Sa., 16. Nov. 2019 08:57:

bleehome:
> Yikes, I had even forgotten about this... So @bleehome is this consistently never working the first time? Would it be possible for you to put together a small sample test to reproduce this?
I made a very simple  demo but it hangs on waiting for ReportedCursorPosition. so f3 works every first time.
[...]

PR #434, opened 2019-11-17
Workaround for possible F3 vs CursorLocation ambiguity.

avl42:
F3-key typically sends `^[OR` if unmodified, or `^[[1;<n>R` with `<n>` between 2 and 8, if shift'ed, ctrl'ed or alt'ed. A CursorLocation on the other hand always looks like `^[[<row>;<col>R`.

Therefore, shift'ed, alt'ed or ctrl'ed F3 can not be safely distinguished from a CursorLocation, but a bare F3 usually can.

This patch will make sure that a bare F3 key cannot get mistakenly taken for a CursorLocation.


avl42:
Suggested fix for Issue #430

Issue #431, opened 2019-10-28
Double With Character Delta Refresh Render dots on the screen


bleehome:
Hi, I use lanterna and MINA sshd to create a sshd server, there is a problem on handling double witdh characters:
The ssh client is Xshell.
For example, there is a Chinese character on position [row=10, col=10], then I put a  empty  space character ' ' to  [row=10, col=10] and invoke a delta refresh. It 's render a dot '.' on position [row=10, col=11].
I use a ActionListBox to render many items with Chinese words, It makes the screen looks like this:
![image](https://user-images.githubusercontent.com/1271744/67667452-cf584000-f9a8-11e9-9d6d-a1d846a9645d.png)
I resolved this problem by
 com.googlecode.lanterna.screen.TerminalScreen#refreshByDelta
![image](https://user-images.githubusercontent.com/1271744/67667642-3118aa00-f9a9-11e9-9372-f49d28acb42d.png)

avl42:
I tried to reproduce it, but no dots showed up.
Maybe that is some artefact of Xshell  ?

mabe02:
I'm also having issues reproducing this. @bleehome could you try to create a really simple program to reproduce this? Also, please check out the lanterna code, build everything and run the CJKTerminalTest program which is among the test sources. Once the program is running, press any key (once) and the program will start printing 'X' to the screen to show how the terminal interacts with double-width characters. Here's what it looks like for me, although I'm using the SwingTerminal here and not a real one:
![bild](https://user-images.githubusercontent.com/10136489/68990170-4d668300-0893-11ea-8eca-dfdd6fd0674e.png)

bleehome:
https://github.com/bleehome/cloudbility/blob/master/lanterna-samples.tar.gz
there is a simple as a maven project
run IssueTest.main, then connect with xshell 6, press F10,  there are some dots on the left-bottom

bleehome:
connect with    
ssh -p 8022 user@127.0.0.1 
and type any characters as password.

avl42:
I can't test Xshell here (no Windows on my machines), but I have a suspicion, that when a doublewidth-character gets replaced by a single-width character, then apparently Xshell writes a `.` at the now uncovered position.  Maybe it could help to detect doublewith characters in the frontbuffer and force the next character into the update range.  Maybe that would fix it for Xshell.

I'll try to prepare a patch for @bleehome to try...

bleehome:
[...]

PR #436, opened 2019-11-17
Maybe fix for Issue #431

avl42:
@mabe02: Only merge, if @bleehome approves it...


mabe02:
We got sign-off from @bleehome so I'm merging this

PR #437, opened 2019-12-04
Fixing format by adding new line where needed

gabrielchl:
Was just going through the examples when I found these two lines stuck together, wherein the previous tutorials, it's in two lines.


mabe02:
Thank you!

Issue #421, opened 2019-06-06
lanterna assumes that Cygwin is installed in the root directory


buko:
Running a lanterna program using the Cygwin terminal results in the well-known issue:

```
Exception in thread "main" java.io.IOException: Cannot run program "/bin/stty": CreateProcess error=2, The system cannot find the file specifie
```

The program /bin/stty.exe exists but it won't be there unless Cygwin was installed in the root directory.

## Workaround

You can work around this problem by putting stty.exe where Lanterna expects it. Create the directory `C:\bin\` and copy `stty.exe` to `C:\bin\stty.exe`.

Once the fie has been copied over the lanterna program will work.

## Proposed Solution

lanterna should check the environment variable `CYGWIN_HOME`. If this environment variable is set then lanterna should (1) Automatically use the Cygwin Terminal and (2) Change all paths to be relative to this base directory.

avl42:
The topic of cygwin and /bin/stty has been raised a couple of times in the past.

The problem is, that so far, about all who reported a problem with "/bin/stty and Cygwin" turned out not to have anything to do with cygwin, and just accidentally stumbled into the error, because they started the lanterna app from command line with java.exe (rather than for plain windows correctly with javaw.exe).

If you happen to be the first one who really has cygwin installed, and intends to run a lanterna app in cygwin's terminal, please let us know!  And also let us know if accessing stty that way will *really* result in a working cygwin-terminal application. 

What I also have gathered from different sources is, that it wouldn't even help calling the right stty.exe from within a java application, as the stty.exe would not have access to the terminal running the java application.  My expectation is, that to run Lanterna under cygwin will really require a shell-script that calls  stty.exe, then the Lanterna-app, and finally restore the tty to original settings.  

Lanterna's  CygwinTerminal is in desperate search for someone working with Cygwin and knowing what needs to be done there to let lanterna read single keystrokes in "raw" mode.

buko:
Well I have cygwin installed and I'm hoping to use lanterna and the workaround does fix the problem. I think it would be a good idea for lanterna to use CYGWIN_HOME to locate stty.

> And also let us know if accessing stty that way will really result in a working cygwin-terminal application.

I've noticed other oddities under cygwin, especially when using the GUI components.

> Lanterna's CygwinTerminal is in desperate search for someone working with Cygwin and knowing what needs to be done there to let lanterna read single keystrokes in "raw" mode.

[...]

PR #441, opened 2019-12-27
DefaultTerminalFactory Graal native-image support

### LINKED ISSUE #440: Graal native image support
Still trying to figure this one out. Running native-image with a dependency on AWT/Swing seems to be an issue. See https://github.com/oracle/graal/issues/664 for example.

As an experiment though, I removed the AWT/Swing terminal emulators from lanterna and native-image worked. And the nice thing was that my command line application was noticeably faster to load while also capable of running at a much lower memory footprint. This makes me wonder if it would make sense to split off the terminal emulators into a separate module so that users could use a lanterna-core without any complications.

There might be some other way around all this but I'm still investigating. I'll post an update when I find out more. I have one other idea that I need to try first.

reibitto:
Resolves #440

I was able to workaround this without needing to do any module splitting. The issue is that the runtime checks in `TerminalFactory.createTerminal` isn't enough to tell Graal that AWT/Swing isn't going to be needed. I created a `createHeadlessTerminal` method which can be verified AOT that the emulators aren't going to be used. Using `createHeadlessTerminal` I'm able to use lanterna + native-image without any issues.

I'm not sure how you feel about this solution. It's simple, but maybe adding a new method to TerminalFactory isn't ideal? But I don't see many other alternatives.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/terminal/DefaultTerminalFactory.java
+++ b/src/main/java/com/googlecode/lanterna/terminal/DefaultTerminalFactory.java
@@ -124,7 +124,13 @@ public class DefaultTerminalFactory implements TerminalFactory {
         }
     }
 
-    @Override
+    /**
+     * Instantiates a Terminal according to the factory implementation with the exception that
+     * {@link DefaultTerminalFactory#preferTerminalEmulator} is always ignored. You may want to use this method when
+     * using tools that rely on AOT compilation such as Graal native-image to ensure AWT/Swing code paths are not hit.
+     * @return Terminal implementation
+     * @throws IOException If there was an I/O error with the underlying input/output system
+     */
     public Terminal createHeadlessTerminal() throws IOException {
         // if tty but have no tty, but do have a port, then go telnet:
         if( telnetPort > 0 && System.console() == null) {
--- a/src/main/java/com/googlecode/lanterna/terminal/DefaultTerminalFactory.java
+++ b/src/main/java/com/googlecode/lanterna/terminal/DefaultTerminalFactory.java
@@ -110,16 +110,7 @@ public class DefaultTerminalFactory implements TerminalFactory {
         //       ("because we can" - unless "rather not")
         if (forceTextTerminal || isAwtHeadless() ||
                 (System.console() != null && !preferTerminalEmulator) ) {
-            // if tty but have no tty, but do have a port, then go telnet:
-            if( telnetPort > 0 && System.console() == null) {
-                return createTelnetTerminal();
-            }
-            if(isOperatingSystemWindows()) {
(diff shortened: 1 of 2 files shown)


jrb0001:
Splitting the module and then using SPI to load the implementations sounds like the more flexible option. But I have no idea how Graal would behave with such a setup and I don't trust it too much after reading about this AWT/Swing issue (the fact that java.desktop isn't needed at runtime is already known at compile time and thus Graal should know that as well).

How well does `createHeadlessTerminal(...)` work on Windows? I would expect "not at all" for all three modes: regular classpath (maybe native bindings as a workaround), JPMS and Graal.

reibitto:
Hmm, I'm not sure about Windows. I'll look into it tomorrow. But yeah, you might be right about that.

reibitto:
After _a lot_ of setup work for Windows and Graal native-image and fixing 4 or 5 different issues I came across, I finally ran into this blocker: https://github.com/oracle/graal/issues/1870. So it looks like that particular experiment will have to wait until next year.

Besides Graal though, I ran into the known issues with lanterna and Windows. WindowsTerminal isn't complete and not in a working state currently, which I can understand given this #422. But even the cygwin support is currently not working which I wasn't expecting to be the case. I missed this issue: https://github.com/mabe02/lanterna/issues/421. I ran into the exact same thing with "/bin/stty" not being found.

So looks like I'm all out of options when it comes to Windows. I guess I can only provide a native executable for Mac and Linux, and continue to rely on the plain JVM + Swing for Windows.

mabe02:
Cygwin support proved to be next to impossible unfortunately, I'll remove it from master branch

mabe02:
So what's the status on this PR, is it ready to be merged? If it only works on Mac+Linux, that's better than nothing.

reibitto:
It's confirmed to be working on macOS at least. I imagine Linux too since it's the same code path.

The goal is to have some sort of solution for Windows too, but I can see that's going to take a lot of time and effort on both fronts. Firstly, there's not much point pursuing it for now since Graal native-image support on Windows is still experimental. I imagine it could take over a year for it to be in a usable state.

Then there's the lanterna Windows support issue. If cygwin support is being dropped, maybe I'll have to look into the native implementation and get that fully tested and stable. Not sure how much work that will be, but it's definitely something I'm interested in supporting. Depending on how successful my upcoming open source project becomes, I might be able to get some help with that. We'll see. This is still months down the line.

buko:
> Cygwin support proved to be next to impossible unfortunately, I'll remove it from master branch

I wonder why Cygwin support is being removed? We actually just got it working finally here a few months ago.

mabe02:
I decided to leave Cygwin in after all, since it is possible to get it somewhat working.

PR #442, opened 2020-01-06
Add IME support to terminal emulators

### LINKED ISSUE #439: CJK/Input Method Editor support for terminal emulator
I tried inputting Japanese and Korean text into the Swing terminal and neither worked correctly. It looks like input coming from IMEs is not supported currently.

I investigated this and supporting it seems doable using [InputMethodListener](https://docs.oracle.com/javase/8/docs/api/java/awt/Component.html#addInputMethodListener-java.awt.event.InputMethodListener-). The idea is basically to simulate CJK key presses when text is committed with the IME.

This might need a little bit of trickery if you need a JTextComponent to receive these events properly. Such as adding a JTextComponent to the terminal emulator frame but not styling it so that it remains invisible. I did confirm though that the events still come through even if you use `setEditable(false)`, so at least there's that.

@mabe02 Does this sound reasonable? If so, I wouldn't mind giving this a shot.

Other ideas are welcome too. I'm unable to see any other alternatives though.

reibitto:
Resolves #439

I positioned the composition box right below the cursor like so:

ss 2020-01-07 01 17 17

### DIFF
--- a/src/main/java/com/googlecode/lanterna/terminal/swing/TerminalInputMethodRequests.java
+++ b/src/main/java/com/googlecode/lanterna/terminal/swing/TerminalInputMethodRequests.java
@@ -1,0 +1,60 @@
+package com.googlecode.lanterna.terminal.swing;
+
+import com.googlecode.lanterna.TerminalPosition;
+
+import java.awt.*;
+import java.awt.font.TextHitInfo;
+import java.awt.im.InputMethodRequests;
+import java.text.AttributedCharacterIterator;
+
+class TerminalInputMethodRequests implements InputMethodRequests {
+
+    private Component owner;
+    private GraphicalTerminalImplementation terminalImplementation;
+    
+    public TerminalInputMethodRequests(Component owner, GraphicalTerminalImplementation terminalImplementation) {
+        this.owner = owner;
+        this.terminalImplementation = terminalImplementation;
+    }
+    
+    @Override
+    public Rectangle getTextLocation(TextHitInfo offset) {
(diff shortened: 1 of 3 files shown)


reibitto:
This technique seems to work quite well for Japanese (and probably Chinese too as the input style is the same). For Korean less so. Inputting itself works, but since most Korean IMEs don't have a composition box, you're typing without feedback until you complete the syllable block _and_ begin the next block. I don't see any alternative with the way things are currently are, but I'll eventually explore some ideas on how to improve the situation for Korean. At least input is possible now.

src/main/java/com/googlecode/lanterna/terminal/swing/TerminalInputMethodRequests.java
```diff
@@ -0,0 +1,60 @@
...
+    
+    public TerminalInputMethodRequests(Component owner, GraphicalTerminalImplementation terminalImplementation) {
+        this.owner = owner;
+        this.terminalImplementation = terminalImplementation;
+    }
+    
+    @Override
+    public Rectangle getTextLocation(TextHitInfo offset) {
+        Point location = owner.getLocationOnScreen();
+        TerminalPosition cursorPosition = terminalImplementation.getCursorPosition();
```
mabe02: This might be null if the cursor is hidden, in which case I think top-left corner is preferred

mabe02: Nevermind, I'm mixing things up. The position should always be valid.

Issue #452, opened 2020-03-26
Triggering button on mouse event


hanek23:
Hi,

When I was trying to enable mouse support in terminal I stumble upon one issue. I can focus all components in my application including buttons but I was not able to trigger them. Then I looked through code and found possibly a simple fix. There is a method in Button (handleKeyStroke) which determines if you have pressed "enter" and if so it triggers all listeners for that button. MouseEvent just goes "unnoticed" to the abstract parent. 

I suggest adding two conditions to that method. One being "OR keystroke is MouseEvent" other one being "AND button.isFocused()". That would cause first mouse event to pass through since the button is not yet focused (that is what the abstract parent will do) but the second one would actually trigger the button, which is what I believe is desirable.

Having said that I know I can just override that function and do it myself but I believe that should be default behavior. But since I do not see the whole picture of this library I decided to talk to you guys first.

Regards,
Petr

hanek23:
Just realized that CheckBox also has its own "handleKeyStroke" and it can probably use this improvement as well.

avl42:
I also think, the way forward is to make standard components deal with mouse-events in a sensible way.

There are just a couple of components, and for some of them (e.g. TextBox) mouse-handling is quite non-trivial.

If you could create a PR that would make most (if not all) of lanterna's components mouse-aware, I think that would be quite a welcome contribution.

E.g. TextBox'es could set the cursor to nearest character, and of course Buttons would trigger.

I like your approach that first click would just focus and only second click would do specific action, but it might be made "customizable" by setting a flag and depending on that flag either limit action to "already focused"-case or "do right-away, regardless of focus" or "set focus and do right-away".

hanek23:
Yeah, I have already ran into a problem with ActionListBox which holds list of Runnables and did not see any, as you said trivial, way of handling that. But I will look into it more and I'll see what I can do.

mabe02:
Yeah, I've kind of left all the mouse event handling out so far, because if done to one of the standard components we should really do it for all. But yes, adding mouse handling would be nice.

hanek23:
[...]

PR #457, opened 2020-04-18
Mouse support, Issue #452

hanek23:
Mouse support for: Button, TextBox, CheckBox, ActionListBox, RadioListBox, Menu and Table and a bit of refactoring. 

I mostly edited handleKeyStroke methods of these components but when I came across MenuBar, Menu and MenuItem  I ran into a problem with how components were drawn if there was or was not Menu at the top. What I found out was that if there is a Menu bar, all components would be down by 1 row but their coordinates would stay the same, as if there was no Menu bar and that was messing up mouse clicks, obviously. So what I have done is: edited AbstactComposite#setComponent to take into account that +1 if there is MenuBar present and since EmptyMenuBar is an inner class I had to do a bit of an ugliness. With that I also removed what I believe was another hack in AbstractBasePane where there was something similar with EmptyMenuBar handling. 

I also added getGlobalPosition in Component interface for convenient access to global coordinates of top the left corner of the component.

One component that I was not able to make reactive to mouse is ComboBox because of that inner window which for some reason was not responsive to mouse clicks.

I tested everything on Linux Mint 19.1 inside standard terminal and telnet. I was not able to get it working in putty though. And obviously I don't know all uses of all components so it is possible it is not even working in some cases. 

I have a few more aesthetic modifications in mind, for example extract that same old "if (enter || mouse || character(' ') && isFocused)" in every handleKeyStroke; some "is" methods on KeyStroke like "isEnter" "isMouseClick" and so on, which would in my opinion make it more readable, but that is for future PR, first I want to get this one through.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/AbstractBasePane.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/AbstractBasePane.java
@@ -18,17 +18,18 @@
  */
 package com.googlecode.lanterna.gui2;
 
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 import com.googlecode.lanterna.TerminalPosition;
 import com.googlecode.lanterna.TerminalSize;
 import com.googlecode.lanterna.graphics.Theme;
+import com.googlecode.lanterna.gui2.Interactable.Result;
 import com.googlecode.lanterna.gui2.menu.MenuBar;
 import com.googlecode.lanterna.input.KeyStroke;
 import com.googlecode.lanterna.input.KeyType;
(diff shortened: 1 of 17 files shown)


mabe02:
You seem to be committing mixed tabs and spaces... :(

mabe02:
I can't easily fix the tabs without rebasing, so I'm probably going to merge it as-is. The code looks fine.

mabe02:
Okay, I've merged and then amended the commit to swap tabs->spaces

hanek23:
Yeah I am sorry about that. I have a habit of clicking auto formatting bind all the time and since we do not have the formatter settings it got all messed up, tried to fix most of it in that branch but I guess I did not get everything.

PR #468, opened 2020-05-03
Ensure Table.selectedRow integrity, and update automatically

### LINKED ISSUE #463: Table selectedRow should never be greater than number of rows
This issue is very similar to #461. Right now, code which removes a row from a `Table` also needs to make sure `selectedRow` is still valid, and does not point to a nonexistent row.

keithkml:
Fixes #463


mabe02:
Looks good, thank you!

PR #469, opened 2020-05-03
Add option to hide scrollbars in Table renderer

### LINKED ISSUE #464: Allow disabling scrollbars in Table
We render some small tables, and the scroll bars take up valuable screen real estate. Also, we do not these tables (and their scroll position) are not be directly interactive, so the scrollbar gives a false impression of interactivity.

keithkml:
Fixes #464

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/table/DefaultTableRenderer.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/table/DefaultTableRenderer.java
@@ -135,7 +135,7 @@ public class DefaultTableRenderer<V> implements TableRenderer<V> {
     }
 
     @Override
-    public boolean areScrollBarsHidden() {
+    public boolean isScrollBarsHidden() {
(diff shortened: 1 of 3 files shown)


src/main/java/com/googlecode/lanterna/gui2/table/DefaultTableRenderer.java
```diff
@@ -131,6 +134,16 @@ public synchronized void setExpandableColumns(Collection<Integer> expandableColu
         this.expandableColumns.addAll(expandableColumns);
     }
 
+    @Override
+    public boolean areScrollBarsHidden() {
```
mabe02, line 138: This should be `isScrollBarsHidden` I think

mabe02:
Just need to resolve that conflict and please check my comment on the method name. Otherwise all good.

keithkml:
Thanks! Renamed the method and rebased.

PR #471, opened 2020-05-03
Add option to render table columns all the way to right edge

### LINKED ISSUE #470: Add option for Table to render all the way to right edge (partial column)
Currently, when a `Table`'s contents are too wide to fit horizontally in the parent component, only the columns that fit are rendered at all. It would be nice for our use case if the next column was rendered, but only the part that fit on the screen.

keithkml:
Fixes #470


mabe02:
Looks good, just need to fix that conflict.

keithkml:
Anything else you need me to change?

mabe02:
That's good, thanks!

PR #478, opened 2020-05-12
Allow AbstractTerminal subclasses to manipulate resize listener behavior

keithkml:
The use case motivating this change is that I'd like to force resize listeners to fire, even when the size didn't change, in order to trigger a full `Screen` refresh.


mabe02:
I'd rather do protected getters for these two fields. Does that work for you?

PR #491, opened 2020-06-14
WIP: Lanterna 490 --- ScrollPanel

ginkoblongata:
Creating this PR early before completion in case there is any comments which can gather.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <packaging>jar</packaging>
     <name>Lanterna</name>
     <description>Java library for creating text-based terminal GUIs</description>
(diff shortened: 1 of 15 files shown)


ginkoblongata:
I think I screwed up this PR.

I think that what happened is I branched off master and then rebased onto release/3.1 when I saw your comment of the other one, thinking that the only difference there was the snapshot value.

So since these other files I think or supposed to only be in master, I think I should rebuild this branch and PR without them?

ginkoblongata:
Abandoning this PR, I've created a newer one without other files.

PR #492, opened 2020-06-14
WIP: rough cut of ScrollPanel --- LANTERNA-491

ginkoblongata:
Creating this PR early before completion in case there is any comments which can gather.


ginkoblongata:
testing with this:

mvn clean install -DskipTests ; java -cp "target/lanterna-3.1.0-SNAPSHOT-tests.jar:target/lanterna-3.1.0-SNAPSHOT.jar" -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 com.googlecode.lanterna.issue.Issue490 --mouse-move

src/main/java/com/googlecode/lanterna/TerminalRect.java
```diff
@@ -0,0 +1,125 @@
...
+package com.googlecode.lanterna;
+
+import java.util.Objects;
+
+/**
+ * This class is immutable and cannot change its internal state after creation.
+ *
+ * @author ginkoblongata
+ */
+public class TerminalRect {
```
mabe02, line 28: I think TerminalRectangle would be a better name

src/main/java/com/googlecode/lanterna/TerminalRect.java
```diff
@@ -0,0 +1,125 @@
...
+    
+    // one of the benefits of immutable: ease of usage
+    public final TerminalPosition position;
+    public final TerminalSize size;
+    public final int x;
+    public final int y;
+    public final int width;
+    public final int height;
```
mabe02, line 36: It seems redundant to have both TerminalPosition and TerminalSize and also store x, y, width and height?

src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
```diff
@@ -65,6 +70,114 @@ protected AbstractListBox(TerminalSize size) {
...
+    }
+    
+    @Override
+    public int getVerticalScrollMaximum() {
+        return getItems().size();
+    }
+    
+    @Override
+    public int getVerticalScrollPosition() {
+        return - scrollOffset.getRow();
```
mabe02, line 98: Is the `-` a typo?

src/main/java/com/googlecode/lanterna/gui2/ScrollPanel.java
```diff
@@ -0,0 +1,172 @@
...
+    
+    public void thumbMouseDrag(boolean isVertical, TerminalPosition position) {
+        scrollableBox.thumbMouseDrag(isVertical, position);
+    }
+    public void mouseUp() {
+        scrollableBox.mouseUp();
+    }
+    
+    @Override
+    public void invalidate() {
```
mabe02, line 93: Don't think you need to override this, it should work anyway. The panel is marked as invalid in the Panel class, which forces a re-render of the whole thing.

src/main/java/com/googlecode/lanterna/gui2/ScrollPanel.java
```diff
@@ -0,0 +1,172 @@
...
+        // ?
+        super.invalidate();
+        // ?
+        scrollableBox.invalidate();
+    }
+    
+    public class ScrollPanelLayoutManager extends BorderLayout {
```
mabe02, line 100: Does this need to be public? It looks like you wouldn't need to call this outside of ScrollPanel?

src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
ginkoblongata: Not a typo, but redoing this part.
The ScrollableBox interface kinda became too cumbersome, so taking a draft 2 approach.

src/main/java/com/googlecode/lanterna/gui2/ScrollPanel.java
ginkoblongata: Ok, yep works without this method, using existing implementation.

src/main/java/com/googlecode/lanterna/gui2/ScrollPanel.java
ginkoblongata: Will make package protected.

src/main/java/com/googlecode/lanterna/TerminalRect.java
ginkoblongata: Yes, the intention is to make the usability of this component more fluid for the different ways in which it may be used.

src/main/java/com/googlecode/lanterna/TerminalRect.java
ginkoblongata: Will change it.

ginkoblongata:
I rebased this branch onto release/3.1, so now, not sure why but github shows those changed file too in the above "Files Changed" link, (showing them as green here even though they are already in the release/3.1 branch)

Here is a link for convenience which shows the actual diff:

[](https://github.com/mabe02/lanterna/compare/release/3.1...ginkoblongata:LANTERNA-491)

ginkoblongata:
testing with this: 

`$ mvn clean install ; java -cp "target/lanterna-3.1.0-SNAPSHOT-tests.jar:target/lanterna-3.1.0-SNAPSHOT.jar" -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 com.googlecode.lanterna.gui2.ImageComponentTest --mouse-move`

ginkoblongata:
This is a breaking change:

https://github.com/mabe02/lanterna/pull/492/commits/fc3c0e1dcbf50aca36fb501abcf99b69287411a5

Removed the goofy foot method setCaretPosition(int line, int column), replaced with setCaretPosition(TerminalPosition position).

ginkoblongata:
I guess the activity around this pull request may be settling, I am not aware something further for it.

avl42:
I couldn't follow this thread in detail, but the topic "scrolling" triggers my memory about one of my past contributions: TerminalScreen's "scrollLines" method, which makes scrolling-by-lines more efficient, if the scrolling area covers most of the terminal width.   So, if the new ScrollArea covers more than 2/3s of terminal width, then using ScrollLines will reduce refresh-flicker noticeably.

If you do not already use it, then I'd strongly recommend doing so, otherwise, all the better :-)

ginkoblongata:
If I understood, some terminals implement scrolling with some kind of control characters or something?

In those cases, do the scroll have to happen for the entire line?  I see that things on the Lanterna side can scroll and be arbitrary width, but once they hit the terminal, do those scrollable Terminals support arbitrary regions of scrolling?
Also, looks like the scrolling would be vertically but is there horizontally also?
Thanks

ginkoblongata:
testing with these:

`$ mvn clean install ; java -cp "target/lanterna-3.1.0-SNAPSHOT-tests.jar:target/lanterna-3.1.0-SNAPSHOT.jar" -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 com.googlecode.lanterna.gui2.SplitPanelTest --mouse-move
`
`$ mvn clean install ; java -cp "target/lanterna-3.1.0-SNAPSHOT-tests.jar:target/lanterna-3.1.0-SNAPSHOT.jar" -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=5005 com.googlecode.lanterna.gui2.ThemeTest --mouse-move
`

ginkoblongata:
Looks good to me in Xterm.

avl42:
Yes, Terminals usually support an escape sequence, that will restrict
scrolling to a given range of lines and other sequences to insert and
delete lines.  I do not know of any sequences to support horizontal
scrolling.  It's probably a much rarer usecase in practice than that of
vertical scrolling of whole lines, anyway ;-)

Lanterna uses a back-buffer and a front-buffer to optimize screen
redrawing: the back-buffer contains what *should* be on screen, and the
front-buffer represents what lanterna thinks, that the terminal actually
looks like.

If we now assume (for the sake of demonstration) that the terminal shows
two panels side by side, where panel A shows lines of letters ("aaa...",
"bbb...",...) and covers more than 3/4 of total width, whereas panel B
shows lines of digits ("111...","222...",...) and let there also be
window-borders that are the same char for the total height of the panels.
Finally lets assume, that some user-interaction now causes panel A to
scroll down by one line (in the sense of bringing into view the next line
below current view - thus moving lines up).

So far, Lanterna would let the scrolled components draw themselves into the
backbuffer, then determine *lots of differences* to the front-buffer, and
redraw the whole screen.

Using "scrollLines", Lanterna would "know" that the application scrolled
the lines covered by panel A, and during redraw it will: 1.) send the
escapes to the terminal to perform the scrolling, 2.) scroll the
front-buffer accordingly, 3) notice that the differences are only 1/4 of
the screen, do an refreshByDelta for the stuff in panel B, which got
previously scrolled along.  (The borders themselves did almost not visibly
change; only the bottom line of the borders will need to be redrawn.)

From this it becomes clear, that the scrollLines-approach is only helpful,
if a certain percentage of the horizontal estate is actually being scrolled.

My implementation only covered Screen, but probably could be extended: We
would need to extract ScrollHint-Management from TerminalScreen.scrollLines
into a separate method, and for a Window or Component, implement
scrollLines such, that if the element covers at least some threshold of
horizontal estate on screen, then it would set scrollHints on the Screen
(though *without* actually scrolling the screen!).

The result should be: if a *sufficiently horizontally-large* Component
scrolls, then the Screen's ScrollHint shall be updated, such that the next
refresh will be more efficient than current redraw-it-all.

PS: my scrollLines implementation was triggered by someone complaining that
his application was too slow/flickery when scrolling.

ginkoblongata:
Thanks for the explanation, good to know.

ginkoblongata:
I think some layer in between the Component and the Screen can do some napkin math and determine on the fly if it would invoke the scrollLines() for the blit. At the Component level, I think it makes sense to render the entire component in the general case and some components may attempt to help with getting to that call quicker, but at the application development time because generally the Component may be arbitrarily positioned and occluded on screen at any particular moment.

ginkoblongata:
Ok, looks like there is now still one thing undone with the pull request.

Now the popup menu windows are 'dismissable' when clicking outside of them, or that is what the user may be led to think, but really that is just an artifact of the clicking bringing up another window over top of the popup.

If the user then drags the window away, they can reveal the unspent popup menu.

I guess I will have to look into that as part of this pull request.

ginkoblongata:
Ok, popups menu's now close when user clicks outside their bounds.

src/main/java/com/googlecode/lanterna/gui2/ImageComponent.java
```diff
@@ -0,0 +1,54 @@
...
+import com.googlecode.lanterna.*;
+import com.googlecode.lanterna.graphics.*;
+
+/**
+ * 
+ * @author ginkoblongata
+ */
+public class ImageComponent extends AbstractInteractableComponent {
+    
+    TextImage textImage;
```
mabe02, line 12: You probably meant this should be `private`

src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
```diff
@@ -535,7 +637,8 @@ public String getLabel(T listBox, int index, V item) {
          */
         public void drawItem(TextGUIGraphics graphics, T listBox, int index, V item, boolean selected, boolean focused) {
             ThemeDefinition themeDefinition = listBox.getTheme().getDefinition(AbstractListBox.class);
-            if(selected && focused) {
+            //if(selected && focused) {
+            if (selected) {
```
mabe02, line 641: Was this change intentional?

src/main/java/com/googlecode/lanterna/gui2/AbstractWindow.java
```diff
@@ -149,14 +149,30 @@ public TerminalPosition toGlobal(TerminalPosition localPosition) {
...
         if(globalPosition == null || lastKnownPosition == null) {
             return null;
         }
         return globalPosition.withRelative(
                 -lastKnownPosition.getColumn() - contentOffset.getColumn(),
                 -lastKnownPosition.getRow() - contentOffset.getRow());
     }
+    
+    @Override
+    public TerminalPosition fromDecoratedGlobal(TerminalPosition globalPosition) {
```
mabe02, line 168: Are you calling this method anywhere? I don't see any usages of it

mabe02:
Ok, so I've reviewed this again and the big issue now is that this PR contains multiple unrelated changes. Can you please rebase this on release/3.1 and only include the relevant changes for ScrollPanel? The other changes should go in as separate PRs, if you want to push them too.

src/main/java/com/googlecode/lanterna/gui2/AbstractListBox.java
ginkoblongata: Yes, it is so that the ListBox item under the mouse is highlighted, it would be the one 'activated' when clicked.  However due to another change in this PR it remains dormant still unfortunately.
Since now events are routed to only the active window.

src/main/java/com/googlecode/lanterna/gui2/ImageComponent.java
ginkoblongata: Yep, changed to private.

src/main/java/com/googlecode/lanterna/gui2/AbstractWindow.java
ginkoblongata: Yep, MultiWindowTextGUI for determination of which window is clicked on.

ginkoblongata:
I think it makes sense to break out some of the unrelated changes and make PR's for those and get those through the pipeline before revisiting this PR.
Then this PR can be revisited after those are in release/3.1 and then this PR can rebase those back in and the diff will be smaller between this branch and release/3.1.

src/main/java/com/googlecode/lanterna/TerminalPosition.java
```diff
@@ -159,6 +159,40 @@ public TerminalPosition with(TerminalPosition position) {
...
+    
+    public TerminalPosition minus(TerminalPosition position) {
+        return withRelative(-position.getColumn(), -position.getRow());
+    }
+    
+    public TerminalPosition multiply(TerminalPosition position) {
+        return new TerminalPosition(column * position.column, row * position.row);
+    }
+    
+    public TerminalPosition divide(TerminalPosition denominator) {
```
mabe02, line 174: I don't see you calling this anywhere? Did you have any particular use case in mind?

src/main/java/com/googlecode/lanterna/gui2/ImageComponent.java
```diff
@@ -0,0 +1,54 @@
...
+            @Override
+            public TerminalPosition getCursorLocation(ImageComponent component) {
+                // when null, lanterna hidden cursor for this component
+                return null;
+            }
+        };
+    }
+    
+    @Override
+    public Result handleKeyStroke(com.googlecode.lanterna.input.KeyStroke keyStroke) {
```
mabe02, line 43: Why is this component interactable? Seems like it shouldn't be, like the existing EmptySpace?

src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
```diff
@@ -46,12 +46,17 @@
     private final WindowManager windowManager;
     private final BasePane backgroundPane;
     private final List<Window> windows;
+    private final List<Window> stableOrderingOfWindows;
```
mabe02, line 49: I'm still not sure why you need this? The `window` list should contain all the windows ordered by Z-index.

src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
```diff
@@ -356,6 +362,9 @@ public synchronized Interactable getFocusedInteractable() {
 
     @Override
     public synchronized boolean handleInput(KeyStroke keyStroke) {
+        ifMouseDownPossiblyChangeActiveWindow(keyStroke);
+        ifMouseDownPossiblyStartTitleDrag(keyStroke);
+        ifMouseDragPossiblyMoveWindow(keyStroke);
```
mabe02, line 367: Shouldn't you stop event propagation here if any of the above matched the event?

src/main/java/com/googlecode/lanterna/gui2/SplitPanel.java
```diff
@@ -0,0 +1,223 @@
...
+        this.isHorizontal = isHorizontal;
+        thumb = makeThumb();
+        setLayoutManager(new ScrollPanelLayoutManager());
+        setRatio(10, 10);
+        
+        addComponent(a);
+        addComponent(thumb);
+        addComponent(b);
+    }
+    ImageComponent makeThumb() {
```
mabe02, line 64: `private`?

src/main/java/com/googlecode/lanterna/gui2/TextBox.java
```diff
@@ -314,31 +368,32 @@ public TerminalPosition getCaretPosition() {
...
     /**
      * Moves the text caret position to a new position in the {@link TextBox}. For single-line {@link TextBox}:es, the
      * line component is not used. If one of the positions are out of bounds, it is automatically set back into range.
      * @param line Which line inside the {@link TextBox} to move the caret to (0 being the first line), ignored if the
      *             {@link TextBox} is single-line
-     * @param column  What column on the specified line to move the text caret to (0 being the first column)
+     * @param position What row & column to move the text caret to (0 being the first column)
      * @return Itself
      */
-    public synchronized TextBox setCaretPosition(int line, int column) {
```
mabe02, line 328: This is a breaking change, please keep the original signature (you can forward the call into your new signature)

src/main/java/com/googlecode/lanterna/TerminalPosition.java
ginkoblongata: I am thinking this was for set completeness.

src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
ginkoblongata: Let's focus on the other pull request and leave this one dormant for now.

src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
ginkoblongata: Nope, the window may not be focused, and then it can become focused and handle the event.
For the drag, it could have some logic flags added for that, but not sure the benefit, since the drag will only occur when the mouse event misses all others anyway.  It would produce more code for the same effect.

src/main/java/com/googlecode/lanterna/gui2/SplitPanel.java
ginkoblongata: Let's focus on the other pull request and leave this one dormant for now.

src/main/java/com/googlecode/lanterna/gui2/TextBox.java
ginkoblongata: Will clean this up once revisiting this PR.

src/main/java/com/googlecode/lanterna/gui2/ImageComponent.java
ginkoblongata: Generally for longer term, future version I would try to make the case that there should not be any non interactable components.  It makes the library more complex to have this distinction and I got hung up on it and no doubt the same can happen to other users.
I think some breaking change version should converge interactable down into the Component.
In the meantime, I'd prefer to always extend Interactable to lessen the future work to make such a simplification of the library.

Specifically though, users would like to make a mouse listener and when not interactable, the image component is much less useful.

ginkoblongata:
Just dropping a note here, haven't forgot about this, hopefully some more activity in a couple weeks.

mabe02:
@ginkoblongata would you be opening a new PR or update this one?

ginkoblongata:
I would like to keep this one open, and eventually have it accepted for merge.  But I will make another 1 or 2 slices off I think with new PR.

ginkoblongata:
Just dropping another note, I am still committed to this change.  Other commitments are currently taking priority at this time still.

scolastico:
Would be very interrested to see that feature.

mabe02:
Stale

PR #493, opened 2020-06-14
Lanterna 487 redux mouse move

ginkoblongata:
Keeps mouse & mouse drag delineated correctly.
Mouse move will change selection in ListBox style widgets but not do the 'action' until mouse pressed, or keyboard actioning.
Other widgets like Button, do not need to have it toggling focus, as the mouse passes throughout the ui screens.  (maybe those should too?)


ginkoblongata:
What happened:

Some context: I was using Konsole to develop, when I went to work on ScrollPanel, I realized that Konsole does not generate adequate mouse position actions.
It seems to use about halfway prior and after the character boundary as pixels for the actual character, the result is that it's inadequate for precision clicking.
So I moved over to XTerm, which has good proper bounds on the characters themselves as far as the columns for mouse actions.
Then I see that when first starting the XTerm, it's generating mouse DRAGs until I click somewhere, rather than the mouse MOVE's.
I updated the mouse action pattern matcher to include state and it coerces these DRAG's to MOVES until the state is mouse DOWN, then during that mouse down state, it coerces MOVE's to DRAGs as well although I did not encounter this case.

ginkoblongata:
I believe this pull request is done and ready for scrutiny and merging.

mabe02:
Looks good, thank you!

PR #498, opened 2020-06-28
some base level additions and minor cleanup and window title bar drag ability

ginkoblongata:


ginkoblongata:
This PR represents commonality sliced out of the PR of the ScrollPanel stuff.

Once this makes it through, I'll put in a PR for the ImageComponent, then another PR for the SplitPanel.

I think then the ScrolPanel PR will be hopefully on topic specifically enough.

Thanks!

src/main/java/com/googlecode/lanterna/gui2/AbstractBasePane.java
```diff
@@ -44,6 +45,8 @@
     private boolean strictFocusChange;
     private boolean enableDirectionBasedMovements;
     private Theme theme;
+    
+    private Interactable mouseDownForDrag = null;
```
ginkoblongata, line 49: This makes it so that the component which experienced the mouse down is the component which the mouse drags get routed to.

src/main/java/com/googlecode/lanterna/gui2/AbstractWindow.java
```diff
@@ -149,14 +149,30 @@ public TerminalPosition toGlobal(TerminalPosition localPosition) {
...
+    public TerminalPosition fromContentGlobal(TerminalPosition globalPosition) {
         if(globalPosition == null || lastKnownPosition == null) {
             return null;
         }
         return globalPosition.withRelative(
                 -lastKnownPosition.getColumn() - contentOffset.getColumn(),
                 -lastKnownPosition.getRow() - contentOffset.getRow());
     }
+    
+    @Override
```
ginkoblongata, line 167: Used by MultiWindowTextGUI.

src/main/java/com/googlecode/lanterna/gui2/ComboBox.java
```diff
@@ -558,15 +557,15 @@ private Result handleEditableCBKeyStroke(KeyStroke keyStroke) {
         public PopupWindow() {
             setHints(Arrays.asList(
                     Hint.NO_FOCUS,
-                    Hint.FIXED_POSITION));
+                    Hint.FIXED_POSITION,
+                    Hint.MENU_POPUP));
```
ginkoblongata, line 561: Used by MultiWindowTextGUI.

src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
```diff
@@ -46,12 +46,17 @@
     private final WindowManager windowManager;
     private final BasePane backgroundPane;
     private final List<Window> windows;
+    private final List<Window> stableOrderingOfWindows;
```
ginkoblongata, line 49: The purpose of the list stableOrderingOfWindows vs the windows list is that the WindowManager would prefer to have the same ordering everytime.

For example a 2x2 grid style window manager would like to arrange the windows in position and size and have those remain regardless of which window is active.
Prior to this change, the windows are coming through in different order because the active window is put to the front of the list, this would force most window managers to do this bookkeeping work which seems to involve overriding multiple methods.

So essentially z-order is what the existing windows list does, but most window managers would probably want access to z-ordering only secondarily to their primary ordering for layout purposes.

mabe02: Ok, but I think in that case rather than having two lists, we should create a `WindowList` class or something like that, which can encapsulate the z-index and the stable index. It would make the code interacting with the list(s) of windows in here much easier.

src/main/java/com/googlecode/lanterna/gui2/Window.java
```diff
@@ -180,6 +181,13 @@ public String toString() {
     @Override
     void invalidate();
 
+    
+    default TerminalRectangle getBounds() {
```
mabe02, line 185: Need to add Javadoc to this one and the two below

src/main/java/com/googlecode/lanterna/gui2/Window.java
```diff
@@ -390,7 +398,11 @@ public String toString() {
      * @return The global coordinates expressed as local coordinates
      */
     @Override
+    @Deprecated
```
mabe02, line 401: Add a @deprecated Javadoc note for this and explain what to use instead

mabe02:
Looks like we can merge this with a few tweaks. For some of the bigger changes in the other PR, I think we want to target master (3.2) for this.

src/main/java/com/googlecode/lanterna/gui2/MultiWindowTextGUI.java
ginkoblongata: I've added a WindowList class.

src/main/java/com/googlecode/lanterna/gui2/Window.java
ginkoblongata: I've updated with some javadocs.

src/main/java/com/googlecode/lanterna/gui2/Window.java
ginkoblongata: I've updated with some javadocs.

ginkoblongata:
Ok, I think this is ready for next round.

avl42:
Since which Java version can keyword "default" be used in that context?


ginkoblongata <notifications@github.com> schrieb am Sa., 11. Juli 2020,
20:48:

jrb0001:
Java 8 introduced the default keyword on interface methods.

ginkoblongata:
Yes, also, by adding to the interface like that with the 'default' it makes it so that existing implementations of the interface can continue to work as is.
I tend to use that feature sparingly but it works out well for that.

I am hoping this PR will make it in.
I have a bunch of thoughts of potential rearrangement type breaking changes for Lanterna but I'd like to keep with the non breaking addition type stuff for now before focusing on that discussion.  I think I've been leaning towards brevity for these but could go into more full context details.

mabe02:
Alright, that is now merged into 3.1

PR #508, opened 2020-09-19
Feature/adding emoji support

mabe02:


src/main/java/com/googlecode/lanterna/TextCharacter.java
```diff
@@ -288,7 +388,10 @@ public TextCharacter withoutModifier(SGR modifier) {
     }
 
     public boolean isDoubleWidth() {
-        return TerminalTextUtils.isCharDoubleWidth(character);
+        // TODO: make this better to work properly with emoji and other complicated "characters"
```
rednoah, line 391: Character alignment on the terminal is even harder than I thought it was... I never noticed that `你好` aligns with `AbCd` and not `AB`.

rednoah: I'd probably go with "if it's CJK then assumes it's double-width, anything else assume single-width". I'm pretty sure that all the other SE Asian languages will behave similar to Thai (lots of accents and modifiers, but always printed single-width in Terminal).

rednoah: Also, I can confirm that Emoji are double-width in macOS Terminal.

mabe02: Yes, this turned out to be a lot harder than I thought! At least this should improve the situation.

Issue #537, opened 2021-02-27
Regression in input from Telnet terminal


ljosa:
The demo in [TelnetTerminalTest.java](https://github.com/mabe02/lanterna/blob/master/src/test/java/com/googlecode/lanterna/terminal/TelnetTerminalTest.java) no longer works.

Bisecting shows that it broke with commit 04acb75068c6e311018e9c2e667dbd9fce66271b.

After connecting, the telnet connection is lost immediately. Here is the stack trace and output:

```
terminal$ java -cp ~/tmp/lanterna/src/main/java:~/tmp/lanterna/src/test/java com.googlecode.lanterna.terminal.TelnetTerminalTest
Resized to {113x36}
java.net.SocketTimeoutException: Read timed out
        at java.net.SocketInputStream.socketRead0(Native Method)
        at java.net.SocketInputStream.read(SocketInputStream.java:150)
        at java.net.SocketInputStream.read(SocketInputStream.java:121)
        at com.googlecode.lanterna.terminal.ansi.TelnetTerminal$TelnetClientIACFilterer.fillBuffer(TelnetTerminal.java:277)
        at com.googlecode.lanterna.terminal.ansi.TelnetTerminal$TelnetClientIACFilterer.available(TelnetTerminal.java:246)
        at com.googlecode.lanterna.terminal.ansi.TelnetTerminal$TelnetClientIACFilterer.read(TelnetTerminal.java:256)
        at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:284)
        at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
        at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
        at java.io.InputStreamReader.read(InputStreamReader.java:184)
        at java.io.BufferedReader.fill(BufferedReader.java:161)
        at java.io.BufferedReader.read(BufferedReader.java:182)
        at com.googlecode.lanterna.input.InputDecoder.getNextCharacter(InputDecoder.java:144)
        at com.googlecode.lanterna.terminal.ansi.StreamBasedTerminal.readInput(StreamBasedTerminal.java:228)
        at com.googlecode.lanterna.terminal.ansi.StreamBasedTerminal.readInput(StreamBasedTerminal.java:207)
        at com.googlecode.lanterna.terminal.ansi.ANSITerminal.readInput(ANSITerminal.java:287)
[...]

PR #539, opened 2021-04-05
Fix for Issue #537

avl42:
I apologize that this even happened in the first place... pah!  It's eclipse's fault, for apparently working despite the bug  ;-) 

Explanation:

When a timeout is set on a Socket, then by documentation the read(...) call throws an exception if it doesn't receive data in time.
eclipse instead just silently returned 0 bytes without throwing an exception, thus I forgot to add the necessary try-catch block.

For now I cannot try this change with eclipse, so I do hope that eclipse at least won't complain about the catch...


mabe02:
Thanks!

PR #542, opened 2021-04-30
Disable thumb in SplitPanel

ivanyu:
Currently, the thumb in `SplitPanel` gets focus on Tab, but itself doesn't handle Tab key again to pass the focus forward. Same for reverse Tab. This effectively breaks the whole tabbing idea if any `SplitPanel` is present in the window.

To fix this, this change disables the thumb component.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/SplitPanel.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/SplitPanel.java
@@ -111,6 +111,7 @@ public class SplitPanel extends Panel {
                 return Result.HANDLED;
             }
         };
+        imageComponent.setEnabled(false);
         
         return imageComponent;
     }


mabe02:
Thanks!

avl42:
Just wondering, if this is the correct solution.    If the thumb is
"disabled" then I'd presume there isn't any way to actually move
it at all by the user.

I haven't used the SplitPanel yet, so I may be wrong here, but for a "way
forward" I would have expected some keybindings being
added to the "thumb", both, for actually moving it and for tabbing out of
it forward or backwards.  Finally, some new option could be
defined on the SplitPane to disable the thumb in cases where it really
doesn't need to be moveable by the user.

I'm just curious if I'm just talking from my ignorance, or whether it's
just a question of someone starting to implement the fancier approach.

mabe02:
Good question. I didn't actually test it :D

mabe02:
I've fixed a couple of things, but the resize is still a bit weird because it only works by mouse (apparently; I didn't get it to work). Will see if I can fix keyboard-based resizing next.

PR #556, opened 2022-02-06
introduced RadioBoxListItemRenderer.getItemText

Raigedas:
customizable item text instead of toString


src/main/java/com/googlecode/lanterna/gui2/RadioBoxList.java
```diff
@@ -252,13 +252,17 @@ public int getHotSpotPositionOnLine(int selectedIndex) {
             return 1;
         }
 
+        protected String getItemText(RadioBoxList<V> listBox, int index, V item) {
```
mabe02, line 255: We don't need the `RadioBoxList<V> listBox` parameter here

mabe02: Ok, I see why you added it, let's go with this.

Issue #559, opened 2022-04-22
Null pointer exception when Mouse enabled and no window


mgallegosig:
Thanks for the great project. I was exploring it and found this issue. I hope it helps.

version 3.2.0-alpha1 (current code seems to be the same)

Given mouse support is enabled and all windows removed
when click on background 
then Null pointer exception 

java.lang.NullPointerException: Cannot invoke "com.googlecode.lanterna.gui2.Window.getHints()" because "priorActiveWindow" is null
	at com.googlecode.lanterna.gui2.MultiWindowTextGUI.ifMouseDownPossiblyChangeActiveWindow(MultiWindowTextGUI.java:391)
	at com.googlecode.lanterna.gui2.MultiWindowTextGUI.handleInput(MultiWindowTextGUI.java:360)
	at com.googlecode.lanterna.gui2.AbstractTextGUI.processInput(AbstractTextGUI.java:94)
	at com.googlecode.lanterna.gui2.AbstractTextGUIThread.processEventsAndUpdate(AbstractTextGUIThread.java:77)
	at com.googlecode.lanterna.gui2.SeparateTextGUIThread.mainGUILoop(SeparateTextGUIThread.java:120)
	at com.googlecode.lanterna.gui2.SeparateTextGUIThread.access$000(SeparateTextGUIThread.java:45)
	at com.googlecode.lanterna.gui2.SeparateTextGUIThread$1.run(SeparateTextGUIThread.java:56)

PR #560, opened 2022-04-23
Fix for Issue #559

avl42:
Not sure if this meets style guidelines, but it seems to me like a minimal fix for #559 .


mabe02:
Thanks!

PR #562, opened 2022-05-11
Fix: pass correct target focus to previous focus

zyuiop:
Hi,

When using this library, I noticed that when overriding `afterLeaveFocus(direction: Interactable.FocusChangeDirection, nextInFocus: Interactable)`, the `nextInFocus` variable was actually set to the current Interactable and not the target one.

This patch fixes that.

Thank you

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,0 +1,1 @@
+sbt.version=1.6.0
--- a/src/main/java/com/googlecode/lanterna/gui2/AbstractBasePane.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/AbstractBasePane.java
@@ -315,7 +315,7 @@ public abstract class AbstractBasePane<T extends BasePane> implements BasePane {
             return;
         }
         if(focusedInteractable != null) {
-            focusedInteractable.onLeaveFocus(direction, focusedInteractable);
+            focusedInteractable.onLeaveFocus(direction, toFocus);
         }
         Interactable previous = focusedInteractable;
         focusedInteractable = toFocus;


mabe02:
Thank you!

mabe02:
Oops, noticed too late that there's an extra file included by mistake in the PR; I'll remove it

avl42:
Isn't that wrong?

When you set focus on a new component, then the *currently focused*
component ought to receive the "Focus Leave" event...

What problem was this meant to fix?

mabe02 ***@***.***> schrieb am Fr., 14. Juli 2023, 15:16:

zyuiop:
The focused interactable receives the event as you say, but the `nextInFocus` should be the newly focused element, not the current one.

mabe02:
I first thought it was wrong too, but then I re-read it and changed my mind. Let me check again...

mabe02:
Looks correct. Javadoc:
```
    /**
     * Method called when keyboard focus moves away from this component
     * @param direction What direction is focus going in
     * @param nextInFocus Which component is receiving focus next (or {@code null} if none)
     */
    void onLeaveFocus(FocusChangeDirection direction, Interactable nextInFocus);
```
Diff
```
         if(focusedInteractable != null) {
-            focusedInteractable.onLeaveFocus(direction, focusedInteractable);
+            focusedInteractable.onLeaveFocus(direction, toFocus);
         }
```
We were passing in the component itself to nextInFocus, but it should be the component that is taking over focus.

avl42:
Oups, sorry....

I misunderstood the onLeaveFocus-call.

The correction seems correct (now to me, too).

mabe02 ***@***.***> schrieb am Sa., 15. Juli 2023, 14:36:

PR #579, opened 2023-08-25
Update GUIGuideWindows.md to use withBorder

nanodeath:
I'm guessing lanterna 3 changed this API a bit, but...yeah, those static bevel methods don't accept a component as a parameter anymore.


mabe02:
Thank you!

PR #585, opened 2023-11-09
Fix 2 bugs

eshraao:
I have a NullPointerException when I remove an AbstractBorder containing a component that has the focus :
```java
Panel root = new Panel();
Button button = new Button("test", root::removeAllComponents);
button.withBorder(Borders.singleLine()).addTo(root);
setComponent(root);
```

A VerticalLinearLayout displays nothing when the Panel has a row size of 1 and two components are added to it :
```java
Panel root = new Panel();
Panel panel = new Panel().addTo(root);
panel.setPreferredSize(new TerminalSize(10, 1));
new Label("test1").addTo(panel);
new Label("test2").addTo(panel);
```

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/AbstractBorder.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/AbstractBorder.java
@@ -50,23 +50,38 @@ public abstract class AbstractBorder extends AbstractComposite<Border> implement
 
     @Override
     public LayoutData getLayoutData() {
+        if(getComponent() == null) {
+            return super.getLayoutData();
+        }
         return getComponent().getLayoutData();
     }
 
     @Override
     public Border setLayoutData(LayoutData ld) {
-        getComponent().setLayoutData(ld);
+        if(getComponent() == null) {
+            super.setLayoutData(ld);
+        } else {
+            getComponent().setLayoutData(ld);
+        }
         return this;
     }
 
     @Override
     public TerminalPosition toBasePane(TerminalPosition position) {
-        return super.toBasePane(position).withRelative(getWrappedComponentTopLeftOffset());
+        TerminalPosition terminalPosition = super.toBasePane(position);
+        if(terminalPosition == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/googlecode/lanterna/gui2/AbstractBorder.java
```diff
@@ -50,23 +50,38 @@ public Border setSize(TerminalSize size) {
...
+        }
         return this;
     }
 
     @Override
     public TerminalPosition toBasePane(TerminalPosition position) {
-        return super.toBasePane(position).withRelative(getWrappedComponentTopLeftOffset());
+        TerminalPosition terminalPosition = super.toBasePane(position);
+        if(terminalPosition == null) {
+            return null;
```
avl42, line 73: Not sure, if null is best choice here...
A border without content might just return its argument, maybe augmented for it's own border "thickness" (typically 1 or maybe 0).

Given that it was an NPE before, anything else would be ok...

avl42 (approved):

src/main/java/com/googlecode/lanterna/gui2/LinearLayout.java
```diff
@@ -21,10 +21,7 @@
 import com.googlecode.lanterna.TerminalPosition;
 import com.googlecode.lanterna.TerminalSize;
 
-import java.util.ArrayList;
-import java.util.IdentityHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
avl42, line 24: I don't really like the "*" with imports...  they've bitten me in other projects already.

please rather list all the necessary classes...

avl42 (approved):

src/main/java/com/googlecode/lanterna/gui2/LinearLayout.java
eshraao: done

mabe02:
Thank you!

PR #589, opened 2023-12-19
Remove duplicated argument to String.format

spaikmos:
There were seven format items sent to the string formatter where only six are required.  One of them was duplicated, so should be removed.


mabe02:
Thank you!

PR #596, opened 2024-02-20
Issue 595: Avoid relying on WeakRef

mabe02:


Kapral67:
Thanks, I proposed some changes to the test in #597

src/main/java/com/googlecode/lanterna/gui2/AnimatedLabel.java
```diff
@@ -164,9 +164,8 @@ private AnimationTimerTask(AnimatedLabel label) {
         @Override
         public void run() {
             AnimatedLabel animatedLabel = labelRef.get();
-            if(animatedLabel == null) {
-                cancel();
-                canCloseTimer();
+            if(animatedLabel == null || animatedLabel.getTextGUI() == null) {
```
Kapral67, line 167: If WaitingDialog::createDialog and WaitingDialog::showDialog have sufficient time between these two calls, the AnimatedLabel will never begin as its textGui is null when the AnimatedLabel starts

Issue #595, opened 2024-02-11
AnimatedLabel Thread hangs after WaitingDialog closed


Kapral67:
It is possible for the `AnimatedLabel::createSpinningLine` component's thread to hang after the `WaitingDialog` that owns it is closed.

A quick fix could be to explicitly call `AnimatedLabel::stopAnimation` in `WaitingDIalog::close` (the spinning line would need to become an instance variable of the `WaitingDialog` class).

Although, I think a better solution would be some fix within the `AnimatedLabel` class itself.

Proof of Concept:
```java
final DefaultTerminalFactory terminalFactory = new DefaultTerminalFactory();
try (final Screen screen = terminalFactory.createScreen()) {
    screen.startScreen();

    // POC
    final WindowBasedTextGUI textGUI = new MultiWindowTextGUI(screen);
    final WaitingDialog waitingDialog = WaitingDialog.createDialog("TITLE", "TEXT");
    waitingDialog.showDialog(textGUI, false);
    CompletableFuture.runAsync(() -> {
                         try {
                             Thread.sleep(5000);
                         } catch (final InterruptedException e) {
                             Thread.currentThread().interrupt();
                             throw new RuntimeException(e);
                         } finally {
                             waitingDialog.close();
                         }
                     }, Executors.newSingleThreadExecutor())
                     .exceptionally(e -> {
                         throw new RuntimeException(e);
                     });
    waitingDialog.waitUntilClosed();
    System.out.println("WAIT DIALOG CLOSED");

} catch (final IOException e) {
    throw new RuntimeException(e);
}
```

mabe02:
Interesting, looks like the WeakReference that's supposed to shut down the thread doesn't trigger until you force GC after the window is closed. We should probably improve on that.

avl42:
Probably, the thread just ought to check validity of the objects it tries
to deal with...
and stop itself, if its prerequisites aren't met.

One should NEVER let program logic depend on actual GC runs.

mabe02:
I think this will fix it:
https://github.com/mabe02/lanterna/pull/596

TheFunnyBear:
Workaround for me:
```

        private fun waitDialog(
            textGUI: WindowBasedTextGUI,
            action: Runnable
        ) {
            val waitDialog = WaitingDialog.createDialog("TETST", "TEST MESSAGE")

            runBlocking {
                val job = CoroutineScope(Dispatchers.IO).launch {
                    action.run()
[...]

PR #597, opened 2024-02-20
Prevent Test From Hanging With AnimatedLabel Thread

Kapral67:
The fix for #595 as proposed by #596 looks good to me.

The test proposed in #596 would hang with the `AnimatedLabel` Thread.

This PR proposes a modification so that now matter if `Issue595` test fails or succeeds:

- it never hangs

- indicates test failure via:

  - output to stderr

  - returned exit code

### DIFF
--- a/src/test/java/com/googlecode/lanterna/issue/Issue595.java
+++ b/src/test/java/com/googlecode/lanterna/issue/Issue595.java
@@ -50,6 +50,7 @@ class Issue595 {
             // POC
             final WaitingDialog waitingDialog = WaitingDialog.createDialog("TITLE", "TEXT");
             final ExecutorService executorService = Executors.newSingleThreadExecutor();
+            Thread.sleep(SLEEP_MILLIS);
             waitingDialog.showDialog(textGUI, false);
             CompletableFuture.runAsync(() -> {
                                  try {
--- a/src/test/java/com/googlecode/lanterna/issue/Issue595.java
+++ b/src/test/java/com/googlecode/lanterna/issue/Issue595.java
@@ -18,6 +18,7 @@
  */
 package com.googlecode.lanterna.issue;
 
+import com.googlecode.lanterna.gui2.AnimatedLabel;
 import com.googlecode.lanterna.gui2.MultiWindowTextGUI;
 import com.googlecode.lanterna.gui2.WindowBasedTextGUI;
 import com.googlecode.lanterna.gui2.dialogs.WaitingDialog;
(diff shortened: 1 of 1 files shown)


src/test/java/com/googlecode/lanterna/issue/Issue595.java
```diff
@@ -18,46 +18,91 @@
...
+            }
+
+        } catch (final Throwable e) {
+            if (e instanceof InterruptedException) {
+                Thread.currentThread()
+                      .interrupt();
+            }
+            ++exit_code;
+            final StringWriter stringWriter = new StringWriter();
+            e.printStackTrace(new PrintWriter(stringWriter));
```
Kapral67, line 102: redundant, can just call `e.printStackTrace()`

avl42: Things like `e.printStackTrace()` are often barfed at by code quality tools, like e.g. SonarQube.
The preferred way would be resorting to some logging-framework, so that it can be disabled
for production use - and not dump an exception-stackdump into user's face.

I'm aware that using a logging framework is not a simple decision, but the original code is already closer to a real solution than "simplifying" it back to `e.printStackTrace()`.

My opinion: just leave that bit as it is.  Simplifying it as suggested would be a step in the wrong direction.

Kapral67: Unless we resort to a logging framework

```java
StringWriter s = new StringWriter();
e.printStackTrace(new PrintWriter(s));
System.err.print(s);
```

Is equivalent to `e.printStackTrace()` [(at least in vanilla JDK implementations)](https://github.com/openjdk/jdk/blob/998d0baab0fd051c38d9fd6021628eb863b80554/src/java.base/share/classes/java/lang/Throwable.java#L677-L679)

avl42:
Yes, they are functionally equivalent.  And, if ever a logging framework is
added, it would be changed
to passing the exception to the logger directly, and not stringifying
beforehand.
(That was originally a thinko on my side, that the logger would then still
need the stringified exception.)

In the meantime, anyone could replace the "System.err" part by some other
channel opened
to some logfile, if he doesn't want exception stack-traces to go in user's
face.  (if terminal is in
raw mode, those exceptions often appear ugly garbled, as the "\n" then is
not mapped to a
linebreak on terminal.

If we changed it to the compact form, replacing the channel becomes more
tedious.

I don't see enough reason to compactify the code towards less flexibility.
Afterall the
two variants under discussion are equivalent at runtime.

src/test/java/com/googlecode/lanterna/issue/Issue595.java
mabe02: This was copy+pasted from the bug report, I think :)

mabe02: Nevermind, wrong PR

mabe02:
Given that this is a test for reproducing a reported bug, I don't really mind that the code style and some constructs are misaligned with the rest of the library. The code won't be included in the library anyway. I've intentionally avoided introducing or using any logging framework in the library itself since there never really was any strong need for logging anywhere. Rather, errors are thrown out and users should catch exceptions and log them as per their application.

PR #601, opened 2024-03-09
TestTerminalSize & TestTerminalPosition and changes to those classes

ginkoblongata:
I can add more comments to this merge request.

Primary change is adding TestTerminalSize and TestTerminalPosition..

Cleaning up TerminalSize and TerminalPosition and getting the memory (same instantce) thing working across the various methods.

Using TerminalSize.of() TerminalPosition.of() rather than constructor to allow for further memory stuff like small pool as example.

### DIFF
--- a/src/examples/java/com/googlecode/lanterna/examples/DrawRectangle.java
+++ b/src/examples/java/com/googlecode/lanterna/examples/DrawRectangle.java
@@ -29,8 +29,7 @@ public class DrawRectangle {
 		screen.startScreen();
 		screen.clear();
 
-		tGraphics.drawRectangle(
-			new TerminalPosition(3,3), new TerminalSize(10,10), '*');
+		tGraphics.drawRectangle(TerminalPosition.of(3,3), TerminalSize.of(10,10), '*');
 		screen.refresh();
 
 		screen.readInput();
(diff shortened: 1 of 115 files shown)


ginkoblongata:
Let me know of anything I need to address for this merge request.
I think it's pretty good, with the size though I've anticipated a combination of further explanation or changes to do for this merge request would happen.
After this merge request then I've got a fix for SplitPanel in the pipeline, and then I'm ready to finally tackle generalized ScrollPanel capability.

avl42:
I think, this PR is too much at once.

Try to split it up into separate topics that can be individually checked and judged.

e.g. I don't really see a use case for multiply and divide operations for Sizes and Positions.

Removing methods is not to be taken lightly (unless it were about private methods.)
getDimension() may be useful, if you write code that ought to work in either horizonal
or vertical direction - LinearLayout would first come to my mind for this.

PS: I do not have any say here - I'm only just a "contributor".
    I only write this feedback, because 1.) that was my own thought after seeing your PR,
    and 2.) Martin hasn't answered on his own, yet.

avl42:
One more comment about the changes in TerminalSize:

1.)  making the fields public is a "no go" - it contradicts "immutability".

2.) I like the static method "of", but its current implementation does 4 comparisons in the non-trivial cases, and up to 6 comparisons in the semi-trivial cases (with only one component in {0,1}).
I think, it might be better to change it to a nested if-structure roughly like that:
```
if (rows==0) {
    if(columns==0) { return ...} else if (columns==1) { return ...}
else if (rows==1) {
    if(columns==0) { return ...} else if (columns==1) { return ...}
}
```

mabe02:
I'm overall ok with this typo of change (on master, since it breaks backward compatibility); just not sure what prompted you to start refactoring these?

Could you maybe squash the commits as well?

Issue #505, opened 2020-08-22
Feature request: emoji support


atoktoto:
The code `textGraphics.putString(0, 3, "🍕")` results in two question mark characters being displayed in the terminal. At the same time `System.out.println("🍕")` works as intended (at least in terminal emulators supporting emojis: ie. the new Windows Terminal).

Is it possible to support this use-case?
I guess the emoji codepoints do not fit into `char` type that is used in `Terminal.putCharacter` so this would require major changes.

rednoah:
`com.ibm.icu.text.BreakIterator` can be used to iterate grapheme clusters (i.e. single-width user-perceived characters):
```java
import com.ibm.icu.text.BreakIterator;
```
```java
public static List<String> getGraphemeClusters(String self) {
	List<String> characters = new ArrayList<String>(self.length());
	BreakIterator i = BreakIterator.getCharacterInstance();
	i.setText(self);
	for (int begin = 0, end = 0; (end = i.next()) != BreakIterator.DONE; begin = i.current()) {
		characters.add(self.substring(begin, end));
	}
	return characters;
}
```

atoktoto:
Seems certainly doable but would require a significant change (or addition) to the Lanterna interfaces. Currently, `char` and `TextCharacter` seems to be the center of the whole operation. Replacing it with `String` representing a single grapheme cluster seems wasteful (in terms of memory) for the general case and can make it less legible: `void putCharacter(String s)` looks wrong :D

mabe02:
I don't see why emoji wouldn't work with the current system, given that we can do CJK characters just fine. I'll investigate, maybe it's the terminal encoding that needs to be updated.

rednoah:
Here's what I get with `lanterna 3.0.3` for a file name such as `THAIบุฟเฟต์EMOJI👩‍👩‍👦‍👦.txt`:
* Works for Thai
* Does not work for complex Emoji

![Screen Shot 2020-08-30 at 09 54 06](https://user-images.githubusercontent.com/1589068/91650205-eb37e100-eaa6-11ea-88e5-b05aa2d95757.png)

That being said, neither `iTerm` nor `Terminal` render this particular Emoji correctly either:
```
 ls
THAIบุฟเฟต์EMOJI👩?👩?👦?👦.txt
```

rednoah:
[...]

PR #625, opened 2025-01-20
Allow printing emojis in StreamBasedTerminals and TextGraphics 

mcarleio:
This is a short term solution for #505.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/graphics/AbstractTextGraphics.java
+++ b/src/main/java/com/googlecode/lanterna/graphics/AbstractTextGraphics.java
@@ -261,11 +261,14 @@ public abstract class AbstractTextGraphics implements TextGraphics {
     @Override
     public TextGraphics putString(int column, int row, String string) {
         string = prepareStringForPut(column, string);
-        int offset = 0;
-        for(int i = 0; i < string.length(); i++) {
-            char character = string.charAt(i);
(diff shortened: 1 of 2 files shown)


pjlegato:
@mabe02 Is Lanterna still active? I'd really like to be able to use emojis with Lanterna. Does this patch do it?
Thanks

avl42:
I think these two commits do not really solve the problem.

They essentially only address the special case, where TextGraphics is used
directly with a
a StreamBasedTerminal.   That completely leaves out  TextScreen and
TextGui, and it
probably also doesn't cover the SwingTerminal (not 100% sure about that,
though).
You probably wouldn't need lanterna in the first place, if writing strings
to stream-based terminal is all you do in your program.

A real solution would need to extend TextCharacter to be able to store
multiple "char"s
that together would represent a single glyph/emoji.

Anyone can fork lanterna repo (preferably to a new name) and check in all
the
maybe incompatible changes.  If it works, maybe mabe02 would merge the PR
into a new major version, but even if not, then it is still all open source!
People could simply opt to use an "emoji-aware clone" of lanterna.

gabrieljones:
@mcarleio would you be able to merge latest [upstream master](https://github.com/mabe02/lanterna/tree/master) into your [branch](https://github.com/mcarleio/lanterna/tree/feat/emoji)?

mabe02:
Thank you!

PR #629, opened 2025-03-04
Basic Swing/AWT mouse support

flizaga:
This small change add the mouse click and scrolls support for Swing terminal (and I guess for AWT as well, but I haven't tested that case honestly)


flizaga:
I just made a mistake when I submit it, it was going on my master branch, so I closed this PR and create a new one going on te right branch. As it seems that it's not possible to edit the branch after PR submission

PR #630, opened 2025-03-05
ENH: Support to-screen coordinates translation

sdedic:
This PR **enhances** `TextGraphics` and `TextGraphicsWriter` (and its subclasses) in that they can translate their relative `TerminalPosition`s to the absolute `TerminalPosition`s used by the `Screen` object. 

This is useful e.g. for computing the current cursor position in an interactive application that writes to TextGraphics an user prompt and then need to position terminal cursor. Or for many other purposes.

I've added also tests to check this new behaviour.

During test development, I've realized there's a significant delay between test setup and the actual test, caused by 5s terminal wait on size query. I added a possibility to configure this delay using system properties; the system property is then used in Maven surefire configuration for test run.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,17 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <version>3.5.2</version>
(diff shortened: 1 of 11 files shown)


mabe02:
Thank you!

PR #631, opened 2025-03-07
Change mouse capture mode to SGR(1006)

flizaga:
Change the mouse capture mode on Terminal based on ANSITerminal from UTF_8 (1005) to [SGR (1006)](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Extended-coordinates)  and add support for the combination of ctrl/alt/shift + mouse event (as this mode allows it to)

The main issues with mode 1005 are : 

- If the terminal is not in UTF8 there could be issues
- Java InputStreamReader converts (at least in Java 8, may be in higher versions too) some characters to 'REPLACEMENT CHARACTER' (U+FFFD), so we end up with wrong coordinates
- The button release event is not clear and does not tell which button has been released, so we have to assume which one it was

Note : This mode exists since 2012 [Xterm Changelog](https://invisible-island.net/xterm/xterm.log.html#xterm_277) and is widely supported by most SSH clients and terminals

### DIFF
--- a/src/main/java/com/googlecode/lanterna/input/MouseAction.java
+++ b/src/main/java/com/googlecode/lanterna/input/MouseAction.java
@@ -43,6 +43,23 @@ public class MouseAction extends KeyStroke {
     }
 
     /**
+     * Constructs a MouseAction based on an action type, a button and a location on the screen
+     * @param actionType The kind of mouse event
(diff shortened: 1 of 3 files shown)


mabe02:
Thank you!

Issue #650, opened 2025-11-19
MessageDialog is not being constructed with "MODAL" hint


Orbisman:
If a MessageBox is opened over an existing window, the focus should not be removed from the MessageBox until the button Cancel/Ok button is pressed.  

Currently if using a mouse setup, and for an example and simple MessageBox with [Ok] button is displayed, if you click on the background window, the focus changes to the background while the MessageBox is still present.

mabe02:
PR merged, thanks!

PR #651, opened 2025-11-19
Patch for Issue #650 - Fix MODAL issues

Orbisman:
- Add MODAL hint to MessageBoxBuilder
- Added check for 'MODAL' in the MultiWindowTextGUI class
- Add check to see if the component has focus in AbstractInteractableComponent class
- Added setMouseCapture() to the DialogsTextGUIBasicText

To recreate:
1. run the DialogsTextGUIBasicText as app
2. Open any of the dialogs
3. Click anywhere not on the dialog, the focus will be changed to the area clicked on.

To validate:
- rerun but in Step 3.  focus should not leave the dialog until the button is pressed.


mabe02:
Two tests broke after merging this:
Issue452Test.testButton:143
Issue452Test.testCheckBox:125

This is due to the component now requiring focus before it accepts the action. I'm tempted to update the tests to match the new logic...

Issue #655, opened 2025-12-21
SplitPanel - Add arrow keys handling / Not resizing when using mouse


Orbisman:
**Enhancements for the SplitPane:**
- Add the ability to use UP / DOWN arrows keys on a 'Horizontal' SplitPanel when it has focus
- Add the ability to use LEFT / RIGHT arrows keys on a 'Vertical' SplitPanel when it has focus.

Standard keys, TAB / <Shift - TAB> work to navigate

**Fixes:**
- When using mouse, the SplitPanel is not resizing when dragging.
- In a vertical layout, the width of the components are not expanding past their initial sizes

mabe02:
@Orbisman should I go with @ginkoblongata's PR instead?

PR #656, opened 2025-12-21
Patch for Issue #655

Orbisman:
See Issue #655 for details.

- Added keystroke handing for UP/DOWN/LEFT/RIGHT ARROW keys.
- Moved resizing code to private function
- First Mouse click will focus component before allowing it to be dragged
- Used same handleKeystroke() layout as table to be be consistent
- Added call to invalidate() at the end of the resize() function to fix the mouse dragging issue.

Testing requires the mouse capture turned on in the TestBase class, ie.

Screen screen = new TestTerminalFactory(args).setMouseCaptureMode(MouseCaptureMode.CLICK_AUTODETECT).createScreen();

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/SplitPanel.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/SplitPanel.java
@@ -18,11 +18,20 @@
  */
 package com.googlecode.lanterna.gui2;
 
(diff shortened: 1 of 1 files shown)


Orbisman:
Going to pull this back, forgot one fix, will  update Issue and resubmit

PR #665, opened 2026-04-02
Fixed mouse movement handling when the position is greater than 127

panicz:
There is a bug which manifests itself when one tries to use mouse pointer on the right hand side of a high-resolution terminal.
Claude Code came up with a hypothesis that I don't fully understand, and proposed a patch I don't fully understand, which seems to solve the issue.

I'm not claiming that this should be merged, but I'm sharing it for a possible future analysis.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/input/MouseCharacterPattern.java
+++ b/src/main/java/com/googlecode/lanterna/input/MouseCharacterPattern.java
@@ -93,12 +93,12 @@ public class MouseCharacterPattern implements CharacterPattern {
                 } else  {
                     button = 5;
                 }
-            } else if((item & 0x2) != 0) {
-                button = 3;
-            } else if((item & 0x1) != 0) {
-                button = 1;
-            } else if((item & 0x1) == 0) {
-                button = 2;
+            } else if((item & 0x3) == 2) {
+                button = 3;  // right
+            } else if((item & 0x3) == 1) {
+                button = 2;  // middle
+            } else {
+                button = 1;  // left
             }
 
             // Get the modifier keys (it seems that they do not are always reported correctly depending on the terminal)
@@ -127,9 +127,9 @@ public class MouseCharacterPattern implements CharacterPattern {
             // Get the move and drag actions
             if((item & 0x20) != 0)
             {
-                if((item & 0x3) != 0)
+                if((item & 0x3) == 3)
                 {
-                    // In move mode, the bits 0, 1 are set in addition to the 6th bit
+                    // button bits 11 with motion flag = no button held = MOVE
                     actionType=MouseActionType.MOVE;
                     button=0;
                 } else {
--- a/src/main/java/com/googlecode/lanterna/input/MouseTranslatingInputStream.java
+++ b/src/main/java/com/googlecode/lanterna/input/MouseTranslatingInputStream.java
@@ -1,0 +1,148 @@
+/*
+ * X10 mouse fallback for terminals that don't support SGR mode 1006
+ * (e.g. ConPTY on WSL2). Lanterna 3.2+ only parses SGR format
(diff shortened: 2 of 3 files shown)


mabe02:
> Claude Code came up with a hypothesis that I don't fully understand, and proposed a patch I don't fully understand, which seems to solve the issue.

😂

flizaga:
Indeed that is pretty funny to use a pull request for it :D 

But, I can explain what Claude Code suggested here.

So first, the changes on[ MouseCharacterPattern.java](https://github.com/mabe02/lanterna/compare/master...panicz:lanterna:fix-x10-mouse-fallback#diff-4e7b27ba1bcadfcc507e269944015218fb2a04c75efd1a7997dbe98115270d43), are completely unrelated to the issue you are trying to fix. (The changes on this file do basically the same thing, just written in a different way)

As the X10/UTF-8(1005) mode is very messy (many softwares preferred to rely on the SGR(1006) mode), the changes will process the input at byte level instead of char level, then if it detects the X10/UTF-8(1005) mode, it will convert the mouse coordinates and button to SGR(1006) then pass the result to [InputDecoder](https://github.com/panicz/lanterna/blob/88a061dbc6c5ae74b5f64015a6619959471c680b/src/main/java/com/googlecode/lanterna/input/InputDecoder.java).  


And for the record, Here is why it is not working as you expected in the first place : 
We cannot use X10/UTF-8(1005) mode ( https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Extended-coordinates ) in the current implementation of the mouse handling in Lanterna, because we are decoding the input ([InputDecoder](https://github.com/panicz/lanterna/blob/88a061dbc6c5ae74b5f64015a6619959471c680b/src/main/java/com/googlecode/lanterna/input/InputDecoder.java)) with Java's "[InputStreamReader](https://docs.oracle.com/javase/8/docs/api/java/io/InputStreamReader.html)" who expects valid UTF-8 characters. 
Because, the coordinates in this mode, are passed "encoded" as an UTF-8 char. 
But some UTF-8 values (for example between 128 (0x80) and 191(0xBF)) are not valid as a first byte in the UTF-8 standard, these kind of value should be coded on 2 bytes instead of 1.
Depending on the Terminal implementation, some are doing it using two bytes for a char between 128 and 2047), some don't and are just sending a byte with the numeric value of the coordinate. For the ones who don't generate char within the valid UTF-8 ranges will generate char that the "InputStreamReader" will replace by the 'REPLACEMENT CHARACTER' (U+FFFD) as the UTF-8 standard suggest to do, which will end up in coordinates who can end up in the col or row 65533 (0xFFFD) sometimes. 
The [SGR(1006)](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Extended-coordinates) mode however, does not have this issue, and add a feature for the mouse button release event, which was unclear in the 1005 mode. 

That's why many softwares preferred the 1006 mode, which is easier to parse and not prone to coordinate encoding difference.

To be honest, I think it is more up to the Windows terminal (used in WSL2) to support properly the SGR mode, as it is a standard since 2012 as I explained in the [pull request #631](https://github.com/mabe02/lanterna/pull/631). 

If I was Claude code, I would have suggested you to just use a different terminal which would be compliant. It would have saved up you some headaches and Claude code tokens ;)

flizaga:
as a comment if someone want to implement that trick, the current PR does not do it efficiently, because it will add an overhead on each input, whether it is mouse or not, and even when it's mouse, whether it is X10 or not.
So if someone wanna do it, i think it would be clever to have a flag somewhere to apply the filter function only if we are on X10 mode and ideally only on mouse events

mabe02:
I agree, would be good if we can apply this only when X10 mode is enabled

PR #668, opened 2026-06-28
Dead code refactor: Removed depreccated code from Table.java

asshifasultana:
##  Dead Code – Deprecated API Removal

**File:**  
`src/main/java/com/googlecode/lanterna/gui2/table/Table.java`

**Severity:**  
Low

## Description

The `Table` class contains two public methods, `getViewTopRow()` and `setViewTopRow(int)`, that are marked as `@Deprecated`. These methods are no longer maintained and simply act as wrappers around newer APIs that provide the same functionality.

## Why It Is Problematic

Deprecated methods increase the public API surface and contribute to technical debt. Maintaining legacy wrapper methods makes the codebase harder to understand and increases maintenance effort. Since modern replacement methods already exist, keeping the deprecated methods provides little value while increasing the risk of future compatibility issues.

## Evidence

The methods `getViewTopRow()` and `setViewTopRow(int)` are annotated with `@Deprecated` in `Table.java`. The class already provides the recommended replacement methods:

- `getViewTopRow()` → `getFirstViewedRowIndex()`
- `setViewTopRow(int)` → `getRenderer().setViewTopRow(int)`

A project-wide search shows no internal usages of these deprecated methods outside `Table.java`, indicating that they are effectively unused within the codebase.

## Recommended Improvement

Remove the deprecated wrapper methods from `Table.java` and update any remaining external code to use the modern APIs:

- Replace `getViewTopRow()` with `getFirstViewedRowIndex()`.
- Replace `setViewTopRow(int)` with `getRenderer().setViewTopRow(int)`.

If these methods are removed in a future release, the API change should be documented in the project's changelog.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/table/Table.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/table/Table.java
@@ -214,16 +214,6 @@ public class Table<V> extends AbstractInteractableComponent<Table<V>> {
         return visibleRows;
     }
 
-    /**
-     * Returns the index of the row that is currently the first row visible. This is always 0 unless scrolling has been
-     * enabled and either the user or the software (through {@code setViewTopRow(..)}) has scrolled down.
-     * @return Index of the row that is currently the first row visible
-     * @deprecated Use the table renderers method instead
-     */
-    @Deprecated
-    public int getViewTopRow() {
-        return getRenderer().getViewTopRow();
-    }
     
     /**
      * Returns the index of the first row that is currently visible.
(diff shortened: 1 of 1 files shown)


asshifasultana:
It's a humble request to review the PR as soon as possible.  It's a project based task given by the faculty and I have to make reports depending on it.

src/main/java/com/googlecode/lanterna/gui2/table/Table.java
```diff
@@ -242,19 +233,7 @@ public int getLastViewedRowIndex() {
...
-     * @param viewTopRow Index of the row that is currently the first row visible
-     * @return Itself
-     * @deprecated Use the table renderers method instead
-     */
-    @Deprecated
-    public synchronized Table<V> setViewTopRow(int viewTopRow) {
-        getRenderer().setViewTopRow(viewTopRow);
-        return this;
-    }
+    //removed dead code here (setViewTopRow(int))
```
mabe02: Don't think we need this comment

mabe02:
I'm ok to remove these on master, but there's another commit about the clipboard, I think we should do that one in another PR and leave this one for just removing the deprecated code.

src/main/java/com/googlecode/lanterna/gui2/table/Table.java
asshifasultana: removed unnecessary comments and TextBox edits like clipboard and other things. I think now its ok to be merged

mabe02:
Please rebase and squash so this is all in one commit, then also remove all those empty lines you're adding, I don't think you intended to include those in the PR.

asshifasultana:
I've rebased and squashed everything into a single commit, and removed the stray empty lines that had accidentally been included. Should be clean now, let me know if anything else needs adjusting.

mabe02:
Thanks!

PR #671, opened 2026-08-24
`ScrollViewer` with Content-Assisted Scrolling

mstrobel:
I actually wrote this about eight years ago, but never got around to finishing it until your repo caught my eye in my starred list this weekend. If I recall, it's more or less complete. The only issue I could remember was that cursors would continue to paint even when their owner controls were scrolled out of view. I added a fix for that today.

I was able to build a really slick internal tool for inspecting a proprietary shared memory database with Lanterna at my last job, and I wanted to give something back. Hopefully you'll find it useful.

I'm sure there's a bug or two, but give it a go and see what you think. There's an interactive test/demo called `ScrollControllerTest` that lets you toggle content-assisted scrolling. It renders a checkerboard, and depending on whether the assist is enabled, it will either scroll normally by lines/columns or will snap to the squares' edges.

I'd wanted to include a proof-of-concept for hosting your `Table` where it would snap to column/row boundaries when scrolling, but I never got around to it.

### DIFF
--- a/src/main/java/com/googlecode/lanterna/gui2/AbstractBasePane.java
+++ b/src/main/java/com/googlecode/lanterna/gui2/AbstractBasePane.java
@@ -247,9 +247,28 @@ public abstract class AbstractBasePane<T extends BasePane> implements BasePane {
         Interactable previous = focusedInteractable;
         focusedInteractable = toFocus;
         if(toFocus != null) {
+            bringIntoView(toFocus);
             toFocus.onEnterFocus(direction, previous);
         }
         invalidate();
+    }
+
+    private void bringIntoView(final Component c) {
+        if (c == null) {
+            return;
+        }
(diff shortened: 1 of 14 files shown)


MartyIX:
> I'm sure there's a bug or two, but give it a go and see what you think. There's an interactive test/demo called `ScrollControllerTest` that lets you toggle content-assisted scrolling. It renders a checkerboard, and depending on whether the assist is enabled, it will either scroll normally by lines/columns or will snap to the squares' edges.

A screenshot would be nice :)

mstrobel:
> A screenshot would be nice :)

I tried to do a screen recording, but the recorder kept trying to time my keystrokes according to a normal typing cadence, and you ended up not being able to really see any of the scrolling. But here's a still.
scrollviewer

With the checkbox checked, scrolling along either axis scrolls exactly one tile into view. Unchecked, it scrolls by one line/column at a time.

